PR #25, opened 2015-01-04
Added 'kallithea-demo'.

xliiv:
It's a demo version, so there is no time consuming features. It's a "minimum viable product".
If there is a need features such as postgres support, etc. can be added.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,6 +7,7 @@ SERVERS = [
   "consul",
   "docs",
   "downloads",
+  "kallithea",
   "hg",
   "jython-web",
   {:name => "loadbalancer", :box=> "ubuntu/precise64", :codename => "precise", :ports => [20000, 20001, 20002, 20003, 20004, 20005, 20100]},
--- a/docs/services/index.rst
+++ b/docs/services/index.rst
@@ -8,3 +8,4 @@ Services
     discovery
     cdn
     postgresql
+    kallithea
--- a/docs/services/kallithea.rst
+++ b/docs/services/kallithea.rst
@@ -1,0 +1,31 @@
+Kallithea
+=========
+
+Development:
+------------
+
+From psf-salt dir do::
+
+    $ vagrant up salt-master
+    $ vagrant up kallithea
+    $ vagrant ssh kallithea
+    vagrant@kallithea:~$ w3m http://127.0.0.1:80
+
+
+Now you can sign in as a test admin user::
+
+    * User: admin1
+    * password: admin1
+
+
+
+F.A.Q:
+^^^^^^
+
+*How to make it visible outside kallithea machine, I mean from host machine?*
+
+    You should forward port, which means adding this line to VagrantFile::
+
+        s_config.vm.network "forwarded_port", guest: 80, host: 8000
+
+    See more details: https://docs.vagrantup.com/v2/getting-started/networking.html
--- a/pillar/base/firewall/kallithea.sls
+++ b/pillar/base/firewall/kallithea.sls
@@ -1,0 +1,5 @@
+{% include "networking.sls" %}
+
+firewall:
+  http:
+    port: 80
--- a/pillar/dev/roles.sls
+++ b/pillar/dev/roles.sls
@@ -11,6 +11,7 @@ roles:
   consul: "consul.vagrant.psf.io"
   docs: "docs.vagrant.psf.io"
   downloads: "downloads.vagrant.psf.io"
+  kallithea: "kallithea.vagrant.psf.io"
   hg: "hg.vagrant.psf.io"
   jython-web: "jython-web.vagrant.psf.io"
   loadbalancer: "loadbalancer.vagrant.psf.io"
--- a/pillar/dev/secrets/kallithea.sls
+++ b/pillar/dev/secrets/kallithea.sls
@@ -1,0 +1,2 @@
+kallithea:
+    beaker-session-secret: "faked-secret"
--- a/pillar/dev/top.sls
+++ b/pillar/dev/top.sls
@@ -27,6 +27,11 @@ base:
     - firewall.rs-lb-backend
     - groups.downloads
 
+  'kallithea':
+    - match: nodegroup
+    - firewall.kallithea
+    - secrets.kallithea
+
   'hg':
     - match: nodegroup
     - firewall.rs-lb-backend
--- a/salt/kallithea/configs/kallithea-init.conf
+++ b/salt/kallithea/configs/kallithea-init.conf
@@ -1,0 +1,13 @@
+description "kallithea"
+
+start on (filesystem)
+stop on runlevel [016]
+
+respawn
+
+setuid kallithea
+setgid kallithea
+
+env PYTHONPATH=/kallithea/data
+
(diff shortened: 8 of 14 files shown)


README.rst
```diff
@@ -3,3 +3,35 @@ Python Infrastructure
 
 * `Documentation <http://infra.psf.io/>`_
 * `IRC: #python-infra on Freenode <http://webchat.freenode.net?channels=%23python-infra>`_
+
+
+Development:
+------------
```
dstufft: The Kallithea documentation should go in the docs instead of the readme. Probably this can just be moved to ``docs/services/kallithea.rst``.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+## any error reports after application crash                                  ##
+## Additionally those settings will be used by Kallithea mailing system       ##
+################################################################################
+#email_to = admin@localhost
+#error_email_from = paste_error@localhost
+#app_email_from = kallithea-noreply@localhost
+#error_message =
+#email_prefix = [Kallithea]
+
+#smtp_server = mail.server.com
```
dstufft: Do we need a SMTP server for this? We can get one setup for it.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+#use = egg:Paste#http
+## nr of worker threads to spawn
+#threadpool_workers = 5
+## max request before thread respawn
+#threadpool_max_requests = 10
+## option to use threads of process
+#use_threadpool = true
+
+## WAITRESS ##
+use = egg:waitress#main
```
dstufft: This says it's using ``egg:waitress`` but the init file above says gunicorn, which is it?

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+## enable proxy prefix middleware
+#filter-with = proxy-prefix
+
+full_stack = true
+static_files = true
+## Available Languages:
+## de en fr ja pl pt_BR ru zh_CN zh_TW
+lang = en
+cache_dir = %(here)s/data
+index_dir = %(here)s/data/index
```
dstufft: Do these two directories store important information? In other words, should they be backed up?

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+app_instance_uuid = {65f405c2-a984-441b-8610-304eec28bec9}
+
+## cut off limit for large diffs (size in bytes)
+cut_off_limit = 256000
+
+## use cache version of scm repo everywhere
+vcs_full_cache = true
+
+## force https in Kallithea, fixes https redirects, assumes it's always https
+force_https = false
```
dstufft: We'll need to turn this on when we actually deploy it. We only want to deploy new services with HTTPS. We don't need Kallithea to handle HSTS headers though, our load balancers will do that.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+#beaker.session.table_name = db_session
+
+## encrypted cookie client side session, good for many instances ##
+#beaker.session.type = cookie
+
+## file based cookies (default) ##
+#beaker.session.type = file
+
+beaker.session.key = kallithea
+beaker.session.secret = {65f405c2-a984-441b-8610-304eec28bec9}
```
dstufft: Does this need to be actually secret? If so it should use pillar data and have a demo key commited inside of ``pillar/dev/secrets/`` and we'll generate a production secret when we deploy it.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+#beaker.session.encrypt_key = <key_for_encryption>
+#beaker.session.validate_key = <validation_key>
+
+## sets session as invalid if it haven't been accessed for given amount of time
+beaker.session.timeout = 2592000
+beaker.session.httponly = true
+#beaker.session.cookie_path = /<your-prefix>
+
+## uncomment for https secure cookie
+beaker.session.secure = false
```
dstufft: Like the HTTPS rule, this should be set to true as well when we deploy it.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+
+
+################
+### [sentry] ###
+################
+
+## sentry is a alternative open source error aggregator
+## you must install python packages `sentry` and `raven` to enable
+
+sentry.dsn = YOUR_DNS
```
dstufft: We have a sentry account, so we should probably use pillar data to enable sentry in production.

salt/kallithea-demo/configs/production.ini
```diff
@@ -0,0 +1,604 @@
...
+logview.sqlalchemy = #faa
+logview.pylons.templating = #bfb
+logview.pylons.util = #eee
+
+#########################################################
+### DB CONFIGS - EACH DB WILL HAVE IT'S OWN CONFIG    ###
+#########################################################
+
+# SQLITE [default]
+sqlalchemy.db1.url = sqlite:///%(here)s/kallithea.db?timeout=60
```
dstufft: Does important information get saved to the database? If so then we should configure it to use the PSF postgresql cluster.

dstufft:
I think the salt state should probably be moved to just be ``kallithea`` not ``kallithea-demo``.

This also needs setup to use the loadbalancer, but that's easy to do and I can do that for you.

README.rst
xliiv: Done.

salt/kallithea-demo/configs/production.ini
xliiv: To get fully functional Kallithea, I think yes.

salt/kallithea-demo/configs/production.ini
xliiv: I think section "[server:main]" only is used when application is run by:
`paster serve <config-file.ini>`

salt/kallithea-demo/configs/production.ini
xliiv: According to this:
http://pylonsbook.com/en/1.0/exploring-pylons.html?highlight=request#configuration-files
i think it doesn't.
But this one does:
*/kallithea/repos*

salt/kallithea-demo/configs/production.ini
xliiv: Nice :)

salt/kallithea-demo/configs/production.ini
xliiv: Done :)

salt/kallithea-demo/configs/production.ini
xliiv: Should I add section similar to this:
https://github.com/python/psf-salt/pull/25/files#diff-45eb43e3c3a3aa644979437ed9a9d4b1R2
?

salt/kallithea-demo/configs/production.ini
xliiv: It depends. I don't know the purpose of Kallithea. If we deploy it for serious using, postgres obviously should be used. Who decides that?

PR #28, opened 2015-01-12
Expires headers for www.python.org static content

frankwiles:
Add 7 day expire headers to:

    - /m/*
    - /static/*
    - /images/*

### DIFF
--- a/salt/pydotorg/config/pydotorg.nginx.jinja
+++ b/salt/pydotorg/config/pydotorg.nginx.jinja
@@ -250,14 +250,17 @@ server {
 
   location /m/ {
     alias /srv/pydotorg/media/;
+    expires 7d;
   }
 
   location /static/ {
     alias /srv/pydotorg/pythondotorg/static-root/;
+    expires 7d;
   }
 
   location /images/ {
     alias /srv/pydotorg/pythondotorg/static-root/images/;
+    expires 7d;
   }
 
   location / {


dstufft:
Would it make sense to have this explicitly use ``Cache-Control`` instead of the expires directive? I think this would cause things to get cached in Fastly with an old ``Expires`` header. It would also allow us to set that these are public cached items.

frankwiles:
Well Fastly is going to cache the content for 5 minutes right? So 'expires 7d' will really be 7 days - 5 minutes, but yeah Cache-Control would be more correct.  Not sure what you mean by "public cached items" is that a Fastly thing?

dstufft:
It would be cached for 7 days unless you're using ``Surrogate-Control`` headers or you have custom Fastly rules put into place that override the ``Cache-Control`` header. I assumed that there were none of these and pydotorg was just relying on a default cache time.

dstufft:
Oh, and a public cache thing is a ``Cache-Control`` directive. By default browsers and proxies assume that content delivered over TLS is "private", which means that basically only the browser caches will cache it (Fastly doesn't count), the "public" directive just says that the response can be cached in a shared cache (e.g. that the response is public and has no personalized information on it).

frankwiles:
I'm fine with doing Cache-Control, want me to update the PR?

dstufft:
Yea I think so

dstufft:
Taken care of in a59052eaef1d0551ac0d413cc742c1b2b04f5e37

PR #35, opened 2015-01-28
Fix pydotorg user umask

frankwiles:
Fixes generated CSS and JS file permissions.


dstufft:
Does this need to run on every highstate or is there a way we can gate this so it only runs if needed? IOW is there a way we can query this information?

frankwiles:
Sure, while that command is quick it is sorta wasteful.  Changed to using ~/.profile

dstufft:
Deployed.

PR #37, opened 2015-02-11
Move to Django 1.7 migrate changes

frankwiles:
I've been moving Python.org to Django 1.7, this changes how migrations are done.  Moving from pre-1.7 syncdb to post-1.7 migrate.


frankwiles:
Thanks for doing that so quickly you get a :star: !!!

PR #42, opened 2015-03-20
Redirect /newjobs URLs to /jobs

malemburg:
We used to run the jobs app under /newjobs during development
and had the RSS feed URL public for a short while. The redirect
should enable us to remove the second location of the jobs app,
avoiding duplicate URLs for every single job posting.


malemburg:
On 20.03.2015 14:12, Benjamin Peterson wrote:
> Merged #42.

Thanks, Benjamin.

-- 
Marc-Andre Lemburg

PR #48, opened 2015-05-03
Fix old legacy PEP redirects

frankwiles:
Reported https://github.com/python/pythondotorg/issues/741

### DIFF
--- a/salt/pydotorg/config/pydotorg.nginx.jinja
+++ b/salt/pydotorg/config/pydotorg.nginx.jinja
@@ -233,6 +233,10 @@ server {
     return 301 https://pypi.python.org/pypi$1?$args;
   }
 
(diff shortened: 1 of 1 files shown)


salt/pydotorg/config/pydotorg.nginx.jinja
```diff
@@ -233,6 +233,10 @@ server {
     return 301 https://pypi.python.org/pypi$1?$args;
   }
 
+  location ~ ^/peps/pep-(.*)\.html$ {
+    return 301 https://www.python.org/dev/peps/pep-$1/;
+  }
+
   location = /peps/pep-0263.html {
```
gutworth: This block can be deleted.

frankwiles:
Oops I overlooked that.  Fixed.

PR #59, opened 2015-09-07
Disable basic auth on Pycon production site

dchukhin:


ewdurbin:
did we also want to make 2016 the default page?

dchukhin:
@ewdurbin sorry about the delay; yes, that would be good. I've opened pull request #60 for that

PR #69, opened 2015-10-05
Upgrades

dstufft:
* Upgrades salt to the latest version
* Switched to using python/psf-packages controlled packages.
* Cleans up some bitrot.


ewdurbin:
Looks great @dstufft. Two things:
  - can you provide the fingerprint for the deb signing key in docs, and sign it with your key?
  - should we perform an upgrade of the salt minions immediately with this?

dstufft:
Well, this is uploading to packagecloud.io and they manage the signing key. Do you just want that fingerprint put there? Or do you want something else?

ewdurbin:
given that they manage the key, fingerprint being documented is probably useless since i doubt they do any signing/uploading to key servers.

dstufft:
If that isn't acceptable to us, I can investigate something else. I was just using that since it removes the need for us to manage it.

ewdurbin:
i'm fine with packagecloud, i just wasn't aware that they managed the keys exclusively.

PR #76, opened 2015-11-06
Make sure git is installed for dstufft's dotfiles

zware:
Currently bringing up a Vagrant VM fails with 'KeyError: git.latest', this seems to fix it.


zware:
Scratch this, it causes issues with any other sls that includes `git:\n  pkg.installed` (such as pydotorg, pycon, soon-to-be speed-web, ...)

PR #81, opened 2016-04-04
Update users.sls

smontanaro:
Add me (this is my work pub key, will add a home key later)

### DIFF
--- a/pillar/base/users.sls
+++ b/pillar/base/users.sls
@@ -297,3 +297,11 @@ users:
       mail:
         allowed: True
         sudo: True
+  skip:
+    fullname: "Skip Montanaro"
+    ssh_keys:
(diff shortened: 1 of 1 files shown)


benjaminp:
DSA keys are deprecated. Please provide a ed25519 key or RSA key of 2048 bits or more.

smontanaro:
> DSA keys are deprecated. Please provide a ed25519 key or RSA key of 2048 bits or more.

Thanks. I don't know what ed25519 is. I have two RSA pub keys. One
file is 235 bytes long (Oct 2002), the other is 407 bytes long (Nov
2005). (The private keys are different lengths, but once again, the
newer file is bigger. Can I assume the longer one is 2048 bits? How
would I know? I can't remember what I had for lunch Wednesday, let
along the parameters on SSH keys from 10-15 years ago.

Skip

smontanaro:
I generated an ed25519 key on my home computer and updated pillar/base/users.sls. Let me know if this works or not.

PR #83, opened 2016-05-21
Redirect toplevel URIs to Python 3 docs

berkerpeksag:
I saw complaints about this on Reddit and Hacker News before.

Here is the latest one: https://asmeurer.github.io/blog/posts/moving-away-from-python-2/#fn:core

### DIFF
--- a/salt/docs/config/nginx.docs-backend.conf
+++ b/salt/docs/config/nginx.docs-backend.conf
@@ -56,12 +56,12 @@ server {
         return 301 https://$host/devguide/documenting.html#building-the-documentation;
     }
 
-    # Map toplevel URIs to Python 2 docs.
+    # Map toplevel URIs to Python 3 docs.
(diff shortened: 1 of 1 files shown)


salt/docs/config/nginx.docs-backend.conf
```diff
@@ -56,15 +56,15 @@ server {
...
-        return 301 https://$host/2/$1;
+        return 301 https://$host/3/$1;
     }
     location ~ ^/(about|bugs|contents|copyright|download|genindex.*|glossary|index|license|py-modindex|search).html$ {
-        return 301 https://$host/2/$1.html;
+        return 301 https://$host/3/$1.html;
     }
     location ~ ^/(searchindex.js|objects.inv)$ {
-        return 301 https://$host/2/$1;
+        return 301 https://$host/3/$1;
```
benjaminp: These are probably mostly fetched by automated tools and I think should remain pointing to 2.

berkerpeksag: Good point, done. Thanks.

PR #85, opened 2016-05-30
Update speed.python.org description

zware:


zware:
@benjaminp: Thanks for the merge(s)!

PR #93, opened 2017-02-11
Update slack-irc.json.jinja

MarkMangoba:


MarkMangoba:
adding pythondotorg-<-->-#pydotorg  slack / irc bridge

PR #96, opened 2017-03-08
haypo: Change my SSH key

haypo:
Replace my old RSA key with a new ed25519 key.


haypo:
I made a similar change in psf-chef: https://github.com/python/psf-chef/pull/163

PR #105, opened 2017-07-12
Enable a simple read-only FTP server on pythontest

ammaraskar:
This is my first time using salt, though not my first time using an orchestration tool. I tested locally on an ubuntu-14.04 server VM. I'd appreciate any feedback if I made any glaring mistakes.

## Technical Details:
This installs and configures the vsftpd package. The running server is configured to only allow anonymous logins, all write commands are disabled at the ftp dameon level. See the `write_enable=False` option, in addition to setting the ftp folder to be `-w` by all. This makes it so that the entire ftp server is read-only. 

In order to allow ftp through the firewall I've used ports 10090 to 10100, ftp needs to have open incoming ports to allow for data transfers. The alternative to using these fixed ports is to use a kernel module that hooks into iptables and snoops on ftp connections to dynamically allow them through. See here for more details: https://serverfault.com/a/38425

Aside from that the only notable thing this does is change the home of the ftp user to one pointed in the git repo, this makes it so that files can be added to the ftp server automatically as they're pulled from git. Hence, this associated pull request in the pythontestdotnet repo should be pulled before this:  https://github.com/python/pythontestdotnet/pull/1

## Why:
Currently the python test suite uses the debian.org FTP servers for testing, however those will be shutting down soon: https://www.debian.org/News/2017/20170425

See the following bug.python.org threads and github issues:

- https://bugs.python.org/issue30883
- http://bugs.python.org/issue22650
- https://github.com/python/pythondotorg/issues/1069

### DIFF
--- a/salt/pythontest/config/vsftpd.conf
+++ b/salt/pythontest/config/vsftpd.conf
@@ -1,0 +1,51 @@
+ftpd_banner=Welcome to the pythontest FTP server
+
+# Run vsftp as a standalone server
+listen=YES
+
+# Allow anonymous FTP. (This is what the test suite uses)
+anonymous_enable=YES
+
+# Use nobody as the FTP user for least privelege
+ftp_username=nobody
+
+# Use the ftp folder from the git repo as the root
+anon_root=/srv/python-testdata/ftp
+
+# The only local users are the python infra team and they
+# do not use ftp for administration
+local_enable=NO
+
+# This option prevents any writing ftp commands from being issued
+write_enable=NO
+
+# Activate directory messages - messages given to remote users when they
+# go into a certain directory.
+dirmessage_enable=YES
+
+# If enabled, vsftpd will display directory listings with the time
+# in  your  local  time  zone.  The default is to display GMT. The
+# times returned by the MDTM FTP command are also affected by this
+# option.
+use_localtime=YES
+
+# Activate logging of uploads/downloads.
+xferlog_enable=YES
+
+# Make sure PORT transfer connections originate from port 20 (ftp-data).
+connect_from_port_20=YES
+# These ports must be opened in the firewall to enable data transfer
+pasv_enable=YES
+pasv_addr_resolve=YES
+pasv_address=www.pythontest.net
+pasv_max_port=10100
+pasv_min_port=10090
+
+# This option should be the name of a directory which is empty.  Also, the
+# directory should not be writable by the ftp user. This directory is used
+# as a secure chroot() jail at times vsftpd does not require filesystem
+# access.
+secure_chroot_dir=/var/run/vsftpd/empty
+
+# Disable PAM support
+session_support=NO
--- a/salt/pythontest/init.sls
+++ b/salt/pythontest/init.sls
@@ -4,6 +4,18 @@ include:
 git:
   pkg.installed
 
+vsftpd:
+  pkg:
+    - installed
+  service.running:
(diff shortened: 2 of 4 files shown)


salt/pythontest/config/vsftpd.conf
```diff
@@ -0,0 +1,45 @@
+ftpd_banner=Welcome to pythontest FTP server
+
+# Run vsftp as a standalone server
+listen=YES
+
+# Allow anonymous FTP. (This is what the test suite uses)
+anonymous_enable=YES
```
ewdurbin: can we also set an `anon_root` to force anonymous access into the appropriate directory?

salt/pythontest/config/vsftpd.conf
```diff
@@ -0,0 +1,45 @@
...
+pasv_min_port=10090
+
+# This option should be the name of a directory which is empty.  Also, the
+# directory should not be writable by the ftp user. This directory is used
+# as a secure chroot() jail at times vsftpd does not require filesystem
+# access.
+secure_chroot_dir=/var/run/vsftpd/empty
+
+# This string is the name of the PAM service vsftpd will use.
+pam_service_name=vsftpd
```
ewdurbin: can we disable pam completely by setting `session_support` to `NO`, i guess this launches `vsftpd` in a lower privilege mode.

from docs:

> session_support — When enabled, vsftpd attempts to maintain login sessions for each user through Pluggable Authentication Modules (PAM). Refer to Chapter 15 Pluggable Authentication Modules (PAM) for more information about PAM. If session logging is not necessary, disabling this option allows vsftpd to run with less processes and lower privileges.
>
> The default value is YES.

salt/pythontest/init.sls
```diff
@@ -4,6 +4,25 @@ include:
 git:
   pkg.installed
 
+ftp:
```
ewdurbin: if we can change this, a `ftp_username` configuration in `vsftpd.conf` would be nice.

ewdurbin:
As a result of opening up such a protocol, we are seeking a possible donation to support running pythontest.net in a sandboxed infrastructure wholly separate from the PSF's current infra.

Again, I don't see this as a strict requirement of deploying this change. However it would make it much less strenuous to deploy ;)

salt/pythontest/config/vsftpd.conf
ammaraskar: For what its worth I think PAM won't be used at all since we set `local_enable=NO` so it doesn't do any user authentication but I don't mind disabling this just in case.

salt/pythontest/init.sls
ammaraskar: Better yet I'll see if we can make the ftp_username `nobody`, that way its garaunteed to have the least amount of privileges, though I'm not sure how that'll fare considering it has to bind to port 21. I'll test and see if that works, if not, then yeah we can call it `testftp`

salt/pythontest/config/vsftpd.conf
ammaraskar: Sure, I think by default anon_root is whatever the home directory of the ftp user is, which is why I change it in the salt file above.

ewdurbin: won't hurt to be explicit :)

ammaraskar: Good point, done.

Issue #109, opened 2017-08-01
/tmp/build_root/environment/ uses python 2.7


JulienPalard:
The docsbuild venv, from  https://github.com/python/psf-salt/blob/master/salt/docs/init.sls#L38 runs python 2.7.

It's only used to run sphinx (not the docsbuild script itself) and I always (by "mistake") built it using "python3 -m venv" when building locally and it always worked perfectly, so the upgrade looks safe.

However as I almost never used salt, I'm still unsure salt is able to upgrade this cleanly, according to the docs a  `- python: python3` may be enough, if someone knows...

MarkMangoba:
@ewdurbin can you review?  It looks like python: python3 is the answer.

ewdurbin:
yes, that should work but will not recreate the virtualenv.

FWIW the distro we're running for docs will provide us with python 3.4

I'll open a PR with what I think would be a good first step

MarkMangoba:
@ewdurbin I did see adding python3-dev, that makes sense.  Thank you for taking a look!

JulienPalard:
Thanks!

@ewdurbin And thanks for mentionning it's Python 3.4, I just tested and blurb is not compatible with Python 3.4 due to (at least) a `SyntaxError: can use starred expression only as assignment target`.

JulienPalard:
@ewdurbin We tried to make blurb work in Python 3.4 (https://github.com/python/core-workflow/pull/170), but we encontred an old bug that has been fixed now (https://github.com/python/core-workflow/issues/171). So we're unsure what's the easiest way... either try to patch, either accept the bug, either upgrade the machine.

Is there a chance docs.iad1.psf.io can get an upgrade?

ewdurbin:
@JulienPalard we should be able to pull a more recent release via PPA, I'll check into it.

JulienPalard:
Since https://github.com/python/docsbuild-scripts/commit/90ed1d08bbba10c972cfd4d1526f4776c295a13a the python3.6 venv is used and it works, thanks.

`/srv/docsbuild/environment/` can now be dropped.

PR #110, opened 2017-08-04
docs: create a new virtualenv with python3

ewdurbin:
we can migrate scripts over by changing their path, then remove the old virtualenv later.

addresses #109


JulienPalard:
LGTM. Did you manually ran this rule? I already see the venv directory.

ewdurbin:
Yeah, I spot checked it and it went well. Will merge up!

PR #114, opened 2017-08-20
Set x-xss-protection header on pydotorg

berkerpeksag:
I don't know whether https://github.com/python/psf-salt/blob/master/salt/haproxy/config/haproxy.cfg.jinja is a better place to put this so I went with `salt/pydotorg/config/pydotorg.nginx.jinja`.


ewdurbin:
thanks @berkerpeksag, this should be good. I'm not sure if we would want to set it globally yet anyway.

PR #117, opened 2017-09-08
Stop updating dev fixtures for pythondotorg

berkerpeksag:
Generating dev fixtures is being replaced by https://github.com/python/pythondotorg/issues/1055.

Question: I still want to keep the existing fixture available but I don't want it to be regenerated every hour. Do I need to keep the following lines as-is?

https://github.com/python/psf-salt/blob/b68ccd7f4482c93ffd4c89bfb32675847ee2e780/salt/pydotorg/init.sls#L55-L60


ewdurbin:
thanks @berkerpeksag

ewdurbin:
Missed your question.... leaving that line as is is probably a good idea so that the file there can continue be served. If the generated fixtures residing there aren't replaceable we should also ensure that directory is backed up.

PR #119, opened 2017-09-19
Set CSRF_TRUSTED_ORIGINS for staging.p.o

berkerpeksag:
We need to set this on staging.p.o because the is_same_domain() function
returns true only when the hostname of the referer is an exact match of
CSRF_COOKIE_DOMAIN.

In our case it returns false because "staging.python.org" is not an
equivalent of "staging.python.org:9000" and as a result we get a 403 error.

See https://github.com/django/django/blob/e7adad27f30396823f3609fcb8699cefb25278bb/django/middleware/csrf.py#L273
for details.


ewdurbin:
@berkerpeksag thanks! this will render on next run of salt (in about 5 minutes)

berkerpeksag:
@ewdurbin thanks for the super quick merge! :)

PR #123, opened 2017-12-09
FIX 301 to the new devguide.

JulienPalard:
The old configuration was redirecting only for:

curl -I 'https://docs.python.org/devguide/(*)'

See https://bugs.python.org/issue30737


ewdurbin:
Look good! Thanks @JulienPalard.

pushed a few purges to make sure and it's working.

PR #137, opened 2018-05-02
Update staffers, PyPI hosting in overview

brainwane:


docs/overview.rst
```diff
@@ -148,11 +148,13 @@ PyCon
    pycon-tech@python.org.
 
 PyPI
-   The `Python Package Index <https://pypi.python.org/>`_ sees the most load of
-   any PSF service. All of its infrastructure runs on Rackspace configured by
-   `pypi-salt <https://github.com/python/pypi-salt>`_, and it is served over
-   Fastly. The infrastructure is maintained by Ernest W. Durbin, Donald Stufft,
-   and Richard Jones.
+   The `Python Package Index <https://pypi.org/>`_ sees the most load of
+   any PSF service. Its source code is available on `GitHub
```
ewdurbin: Seems the CI check for docs render is angry about

```rst
`GitHub <https://github.com/pypa/warehouse>`_
```

with error:

```
/home/travis/build/python/psf-salt/docs/overview.rst:2:Duplicate explicit target name: "github".
```

Though I'm not sure.

sigmavirus24: docutils only ever wants one piece of text to have an anchor so while you can have as many `[GitHub](...)` in markdown, `reStructuredText` wants each anchor to have a unique display text (for reasons I don't fully understand). There may be another `GitHub` in here in which case you can probably just reuse the other link if it points to the same place by doing

```
 `GitHub`_
```

brainwane: Sorry. Fixed.

brainwane:
Ready for re-review.

PR #138, opened 2018-05-17
Update users.sls

msapiro:
Please add Abhilash Raj who is a mailman developer/maintainer and will be helping maintain Mailman 3 on mail.python.org


ewdurbin:
@maxking, please confirm. I note that the key added matches keys available at https://github.com/maxking.keys but would like your 👍

PR #151, opened 2018-11-02
Docs: Directly jump in venv to get sentry.

JulienPalard:
By just using the global Python we're not having sentry installed.

But using the venv directly is fine: we'll get sentry, and it would simplify passing the venv to subprocesses. Which is supported down to the Makefile since https://github.com/python/cpython/commit/590665c399fc4aa3c4a9f8e7104d43a02e9f3a0c#diff-c8dee9c3c068ad3447bbf459950a35b5 which is nice.


ewdurbin:
good catch!

PR #156, opened 2018-11-23
Docs: Will probably never remove this one.

JulienPalard:
We're having like 100 hits per day on this one according to ```zcat /var/log/nginx/docs-backend.access.log.{2..7}.gz | awk '{if ($9 == 301) {print $7}}' | grep -P '^/((2|3)(\.[0-8])?|dev)/\w+/[\d\w\.]+(?!\.html)$' | wc -l```

Some are bots, some are humans using it as a feature (maybe like 1 of 2).

### DIFF
--- a/salt/docs/config/nginx.docs-backend.conf
+++ b/salt/docs/config/nginx.docs-backend.conf
@@ -75,7 +75,8 @@ server {
         return 301 https://$host/3/$1;
     }
 
-    # Emulate Apache's content-negotiation. Probably should remove eventually.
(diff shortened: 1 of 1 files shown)


Seluj78:
I am one of the users of this feature. I find it really usefull. Why would you be required to add `.html` at the end of the url, seems useless to me and when you forget, that `301` is really usefull

ewdurbin:
Sorry for delay here @JulienPalard, I _think_ I've configured this repo to email me whenever a new PR/Issue is opened via 8553e0e45c27a7b02fc9c7bb7009620e157e5f4e

PR #166, opened 2019-03-10
Setup testing

scottilee:
Adding Docker-based testing with Test Kitchen and two basic tests.

### DIFF
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -3,6 +3,8 @@ Python Infrastructure Tests
 
 Python Infrastructure tests are run on Docker containers with Kitchen Salt (https://kitchen.saltstack.com), testinfra (https://github.com/philpep/testinfra/), and pytest (https://pytest.org).
 
-1. Install Bundler (`gem install bundler`)
-2. Install gems in Gemfile (`bundle install`)
-3. Run `bundle exec kitchen test` to verify tests
+To get started, in the root directory do the following:
+1. Install Bundler (`gem install --user bundler`)
+2. Install gems in Gemfile (`bundle install --path vendor/bundle`)
+3. Run `bundle exec kitchen test` (pytest must be installed)
+  - `bundle exec kitchen test` runs create, converge, verify, and destroy, all of which can be run individually (`bundle exec kitchen create`, `bundle exec kitchen converge`, etc.)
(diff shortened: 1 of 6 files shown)


tests/README.rst
```diff
@@ -0,0 +1,8 @@
+Python Infrastructure Tests
+===========================
+
+Python Infrastructure tests are run on Docker containers with Kitchen Salt (https://kitchen.saltstack.com), testinfra (https://github.com/philpep/testinfra/), and pytest (https://pytest.org).
+
+1. Install Bundler (`gem install bundler`)
```
ewdurbin: Might want to encourage people to use `gem install --user bundler` rather than installing into system? I don't know best practices for Ruby environments

tests/README.rst
```diff
@@ -0,0 +1,8 @@
...
+2. Install gems in Gemfile (`bundle install`)
```
ewdurbin: Similarly I ran into a system vs local issue here and used `bundle install --path vendor/bundle` to keep from installing into system paths.

ewdurbin:
Following the instructions, and with just a few changes to keep from mutating my system Ruby I was able to successfully execute tests! Thank you @scottilee!!

A few open questions I have, mostly as being unfamiliar with kitchen:

1. Is there a way for us to test a given state or set of states in isolation?
2. What if any methods are there for collecting the results?
3. Is there some way to cache the resulting state of the containers after salt is installed? That's currently the slowest part of running the test suite as it installs repeatedly.

This is awesome and a great harness to start building off of if we can address at least 1 and 3 above.

tests/README.rst
scottilee: Fixed.

tests/README.rst
scottilee: Fixed.

scottilee:
1. I think this is what you're talking about, let me know if it isn't: https://github.com/saltstack/kitchen-salt/blob/master/docs/provisioner_options.md#state_collection
2. Something cool I found is that you can send the logs to bpaste.net. To enable this I can add `--pastebin=all` to the pytest command in the suites section of .kitchen.yml. Let me know if you want me to make this the default.
3. I updated the docs but instead of running `bundle exec kitchen test` you can run:
 - `bundle exec kitchen create`
 - `bundle exec kitchen converge`
 - `bundle exec kitchen verify` -- then you can run this command again and again without having to recreate the whole container each time

PR #174, opened 2019-05-16
Add myself to python-speed

pablogsal:
As instructed in:
https://github.com/python/psf-chef/pull/174

CC: @ewdurbin


pablogsal:
Thanks @ewdurbin

PR #176, opened 2019-06-23
SSH pubkey update

patrickbenkoetter:
Removed old RSA pubkey
Added current RSA and EDCSA pubkeys


ewdurbin:
applied. these keys should now be usable.

PR #180, opened 2019-08-01
Allow rouilj access to manage roundup tracker.

rouilj:
We need to remove some spam from the tracker and to obliterate it I need shell access.


ewdurbin:
@rouilj all set.

PR #191, opened 2020-05-23
Doc: Redirection for /id/ and /es/.

JulienPalard:


ewdurbin:
This has been deployed. It may take some time to fully take effect if any of the URLs were cached previously.

PR #197, opened 2020-12-04
implement support for webhook in docsbuild

ewdurbin:
see https://github.com/python/docsbuild-scripts/pull/101


JulienPalard:
I'm not a salt user so I may not be of any good help here. But I merged the server branch, so we can slowly start to test it.

JulienPalard:
@ewdurbin any news for me on this?

PR #198, opened 2021-01-08
Doc: Hello Python 3.9 and 3.10.

JulienPalard:
Related to https://bugs.python.org/issue42869

This should make `https://docs.python.org/3.9/library/functions` redirect like `https://docs.python.org/3.8/library/functions` does.

Currently:
```
$ curl -i https://docs.python.org/3.8/library/functions | head -n 1
HTTP/2 301 

$ curl -i https://docs.python.org/3.9/library/functions | head -n 1
HTTP/2 404 
```

Beware, I did *not* tested it locally.

### DIFF
--- a/salt/docs/config/nginx.docs-backend.conf
+++ b/salt/docs/config/nginx.docs-backend.conf
@@ -85,7 +85,7 @@ server {
 
     # Emulate Apache's content-negotiation. Was a temporary measure,
     # but now people are using it like a feature.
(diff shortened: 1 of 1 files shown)


ewdurbin (approved):

ewdurbin:
merged and in process of being applied.

JulienPalard:
And it works \o/

That was fast, thank you @ewdurbin, it will avoid some people working in 3.9 and using pydoc to find the 404.

PR #199, opened 2021-01-24
Add get-pip.py for Python 3.5

pradyunsg:
Toward https://github.com/pypa/get-pip/issues/83


ewdurbin:
Thanks @pradyunsg, this is now applied and live, but it doesn't appear that the file is available yet.

ewdurbin:
Ope, my bad :) Cache is now cleared and this is all good.

PR #200, opened 2021-01-26
buildbot: Upgrade Python from 3.8 to 3.9

vstinner:


vstinner:
I upgraded the buildbot configuration to replaced Python 3.8 with Python 3.9:
https://github.com/python/buildmaster-config/commit/e0a2b303c180425a759bbd4783945f0cd3e1b94e

I already installed python3.9-dev and python3.9-venv manually. This PR is to update the Salt config.

vstinner:
https://buildbot.python.org/all/#/about says "Python version: 3.9.1".

PR #211, opened 2021-05-26
Move bugs.python.org IRC notifications to Libera Chat

vstinner:


pablogsal (approved):
LGTM

vstinner:
cc @ewdurbin

zware:
LGTM, but we may run into the same SASL issues here as we did with buildbot.  The somewhat hacky solution I have in place there is to run a [znc](https://wiki.znc.in/ZNC) bouncer on the buildbot host.

ewdurbin:
let's give it a shot.

ewdurbin:
> LGTM, but we may run into the same SASL issues here as we did with buildbot. The somewhat hacky solution I have in place there is to run a [znc](https://wiki.znc.in/ZNC) bouncer on the buildbot host.

Is intended to be a durable solution? If so we should actually add it to the buildbot salt configuration. Otherwise it will be lost in the event of a server failure and recovery from backup scenario.

ewdurbin:
Tried to validate that this change did anything but not sure how to 😂

I don't see any errors in irker's log though.

PR #215, opened 2021-06-21
Setup nginx to compress responses from roundup trackers.

rouilj:
Data returned from the roundup servers is not compressed. Set up nginx to compress the data on the fly.

I overrode the default compression level from 1 to 6 as that seems to provide the best compression ratio/cpu time
on the roundup assets on my system. But lowering that is the system is too stressed works too.

I am not sure if there is an additional proxy upstream, so I enabled gzip_vary. It may not be needed.

Made sure to include xml as a compressed type as the RSS feed uses it.


ewdurbin:
I moved the gzip configuration into the TLS server block, and removed `gzip_vary` as nginx is the last proxy before the webserver (gunicorn).

PR #229, opened 2022-02-05
d.p.o/3.7/library/macpath.html does in fact exists.

JulienPalard:
https://docs.python.org/3.7/library/macpath.html redirects, but the file exists:

```
mdk@docs:~$ file /srv/docs.python.org/3.7/library/macpath.html 
/srv/docs.python.org/3.7/library/macpath.html: HTML document, UTF-8 Unicode text, with very long lines
```


ewdurbin:
deployed.

PR #242, opened 2022-06-23
Necessary state changes 

cegerhardson:
These necessary state changes will modify anywhere the roles key from pillar is accessed.

### DIFF
--- a/salt/_extensions/pillar/ca.py
+++ b/salt/_extensions/pillar/ca.py
@@ -12,13 +12,13 @@ import OpenSSL
 
 
 def compound(tgt, minion_id=None):
-    opts = {'grains': __grains__}
+    opts = {"grains": __grains__}
     if minion_id is not None:
         if not isinstance(minion_id, string_types):
             minion_id = str(minion_id)
         else:
-            minion_id = __grains__['id']
-    opts['id'] = minion_id
+            minion_id = __grains__["id"]
+    opts["id"] = minion_id
     matcher = salt.minion.Matcher(opts, __salt__)
     try:
         return matcher.compound_match(tgt)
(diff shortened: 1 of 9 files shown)


salt/_extensions/pillar/ca.py
```diff
@@ -320,7 +320,11 @@ def ext_pillar(minion_id, pillar, base="/etc/ssl", name="PSFCA", cert_opts=None)
     gen_certs = pillar.get("tls", {}).get("gen_certs", {})
     for certificate, config in gen_certs.items():
         role_patterns = [
-            pillar.get("roles", {}).get(r) for r in config.get("roles", "")
+            role.get("pattern")
```
ewdurbin, line 323: This change is necessary due to the new structure of the pillar key.

Note that we need to filter out the `None` values that would result from values that do not define a `pattern` key in their contents... which would ultimately not work anyway. That is why we nest the two list comprehensions and add the if clause.

salt/users/init.sls
```diff
@@ -4,7 +4,7 @@ include:
 {% for user_name, user_config in salt["pillar.get"]("users", {}).items() %}
 {% set admin = user_config.get("admin", false) %}
 {% set access = {} %}
-{% for pat, data in user_config.get("access", {}).items() if salt["match.compound"](salt["pillar.get"]("roles:" + pat)) %}  # " Syntax fix
+{% for pat, data in user_config.get("access", {}).items() if salt["match.compound"](salt["pillar.get"]("roles:" + pat + ":pattern")) %}  # " Syntax fix
```
ewdurbin, line 7: Salt allows for accessing nested keys with a `:` syntax, see https://docs.saltproject.io/en/latest/ref/modules/all/salt.modules.pillar.html#salt.modules.pillar.get.

This updates the `pillar.get` call to access the sub-key we need.

In an explicit example,

going from:

```
roles:
  salt: salt.nyc1.psf.io
```

```
roles:
  salt:
    pattern: salt.nyc1.psf.io
```

We need to go from accessing `roles:salt` to `roles:salt:pattern` which is why the `+ ":pattern"` is added to the string being queried for.

ewdurbin (approved):

PR #245, opened 2022-07-06
Convert local dev to Docker

ewdurbin:
In general this is the only obvious way to get a local setup on M1 Macs now :/


ewdurbin:
We can now cleanly `vagrant up` a docker based salt-master! It's running consul as well, so we can theoretically launch further containers for other roles locally without _having_ to launch consul as well!

PR #252, opened 2022-07-12
Vagrantfile update

cegerhardson:
Update Vagrantfile to correct server list.


Vagrantfile
```diff
@@ -4,18 +4,22 @@
...
+  "mail",
   {:name => "loadbalancer", :ports => [20000, 20001, 20002, 20003, 20004, 20005, 20100]},
   "planet",
   {:name => "postgresql-primary", :roles => ["postgresql", "postgresql-primary"]},
   {:name => "postgresql-replica", :roles => ["postgresql", "postgresql-replica"]},
-  "speed-web",
-  "pypa-web",
+  "pypy-web",
+  "pythontest",
+  "web-pypa",
```
ewdurbin, line 22: This one is a little more confusing. The vagrant name is `pypa-web.vagrant.psf.io` but the role is `web-pypa`. I think this is  the right change for the Vagrantfile to make the hostname match the role, but it will require updating the pattern in dev/pillar/roles.sls

ewdurbin (changes requested):

ewdurbin (approved):

PR #253, opened 2022-07-12
Update vagrantfile

cegerhardson:
Made necessary changes and additions to update the vagrantfile to match correct list of servers.
Also made changes to pillar/dev/roles.sls to match web-pypa's role with its hostname.


cegerhardson:
closed in favor of #252

PR #262, opened 2022-07-28
Update moin configuration to run in development 

cegerhardson:


salt/moin/init.sls
```diff
@@ -2,21 +2,23 @@ moin:
   group.present:
     - system: True
 
-  user.present:
+  user.present:
```
ewdurbin, line 5: we should trim trailing whitespace here.

ewdurbin (approved):
one small nit! otherwise this looks great!

PR #274, opened 2022-08-02
salt-lint so everyone's wearing their sunday's best

cegerhardson:


salt/postgresql/server/wal-e.sls
```diff
@@ -120,14 +120,14 @@ wal-e-install:
...
     - require:
       - pkg: postgresql-server
 
 
 wal-e-gpg-key:
-  cmd.wait:
+  cmd.run:
     - name: gpg --import /var/lib/postgresql/wal-e.gpg
     - user: postgres
     - watch:
```
ewdurbin, line 133: Salts documentation recommends moving from `cmd.wait` with `watch` to `cmd.run` with `onchanges` [here](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait). I think we want to change this `watch` directive to be `onchanges`

salt/consul/init.sls
```diff
@@ -120,7 +120,7 @@ consul:
 consul-template:
   pkg.installed: []
 
-  cmd.wait:
+  cmd.run:
```
ewdurbin, line 123: Salts documentation recommends moving from `cmd.wait` with `watch` to `cmd.run` with `onchanges` [here](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait). I think we want to change the `watch` directive a few lines down to be `onchanges`

ewdurbin (changes requested):
A few notes on the update from `cmd.wait` to `cmd.run` with `onchanges` per Salt's documentation: https://docs.saltproject.io/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait

ewdurbin (approved):

ewdurbin:
Thanks @cegerhardson

PR #292, opened 2023-01-03
Add automation to update Fastly IP list

### LINKED ISSUE #291: Automate fetching lists of new Fastly IPs
#290 updated the firewall for the cdn-logs host with the current ranges of Fastly IP addresses fetched via the API at https://developer.fastly.com/reference/api/utils/public-ip-list/

We should add a mechanism to automatically keep this list up to date so we don't cause logs to stop flowing again. Whatever range was being used to send syslog streams changed sometime in may and it was missed causing fastly to stop reporting any logs at all.

egeakman:
- The workflow runs every Monday at 00:00.
- It can also be run manually through Actions >> Update Fastly IP List (on the left panel) >> Run workflow.

The workflow normally pushes directly to the default branch, but we can configure it to push to another branch and then open a pull request. Also, we can change the file names and their locations if it doesn't fit this project's hierarchy.

Closes #291


egeakman:
CC: @ewdurbin

egeakman:
Closing the pull request since the issue has been resolved. Although it would have been nice if someone had taken a look.
By the way, I am still waiting for the d.p.o logs (https://github.com/python/python-docs-es/issues/2253#issuecomment-1383340956).

PR #294, opened 2023-01-10
edit logrotate.timer from daily to hourly

cegerhardson:


salt/cdn-logs/init.sls
```diff
@@ -14,6 +14,16 @@
   file.managed:
     - source: salt://cdn-logs/config/fastly.logrotate.conf
 
+logrotate_time_hourly:
+  file.replace:
+    - name: /etc/systemd/system/timers.target.wants/logrotate.timer
+    - pattern: OnCalendar=daily
+    - repl: OnCalendar=hourly
+
 /etc/cron.hourly/logrotate:
```
ewdurbin, line 23: I think this state can be removed completely, since the job in question doesn't fire through cron anymore.

salt/cdn-logs/init.sls
```diff
@@ -14,6 +14,16 @@
...
   file.symlink:
     - target: /etc/cron.daily/logrotate
+  cmd.run:
```
ewdurbin, line 26: This command should run in response to the systemd configuration changing, not the cron configuration. I think it should be moved into the state above `logrotate_time_hourly`.

salt/cdn-logs/init.sls
```diff
@@ -14,6 +14,12 @@
   file.managed:
     - source: salt://cdn-logs/config/fastly.logrotate.conf
 
-/etc/cron.hourly/logrotate:
-  file.symlink:
-    - target: /etc/cron.daily/logrotate
+logrotate_time_hourly:
+  file.replace:
+    - name: /etc/systemd/system/timers.target.wants/logrotate.timer
+    - pattern: OnCalendar=daily
```
ewdurbin, line 20: My only concern with this approach is that if the value is changed to anything other than `daily` (manually, or via an update, or .... I dunno 👽), our state will not detect it and set the value we want `hourly`. This is probably OK, but a bit fragile.

Were you able to consider https://docs.saltproject.io/en/latest/ref/states/all/salt.states.ini_manage.html#salt.states.ini_manage.options_present for editing this file?

salt/cdn-logs/init.sls
```diff
@@ -14,6 +14,14 @@
...
-/etc/cron.hourly/logrotate:
-  file.symlink:
-    - target: /etc/cron.daily/logrotate
+/etc/systemd/system/timers.target.wants/logrotate.timer:
+  ini.options_present:
+    - name: /etc/systemd/system/timers.target.wants/logrotate.timer
+    - separator: '='
+    - sections:
+        Timer:
+          OnCalendar: daily
```
ewdurbin, line 23:
```suggestion
          OnCalendar: hourly
```

ewdurbin: I think we want to make sure it's hourly, not daily! Otherwise 💯

salt/cdn-logs/init.sls
```diff
@@ -14,6 +14,14 @@
...
 
-/etc/cron.hourly/logrotate:
-  file.symlink:
-    - target: /etc/cron.daily/logrotate
+/etc/systemd/system/timers.target.wants/logrotate.timer:
+  ini.options_present:
+    - name: /etc/systemd/system/timers.target.wants/logrotate.timer
+    - separator: '='
+    - sections:
+        Timer:
```
ewdurbin, line 22:
```suggestion
        Unit:
          Description: 'Hourly rotation of log files'
        Timer:
```

ewdurbin: just for show, update the description to match what we want!

ewdurbin (changes requested):

ewdurbin (approved):

PR #295, opened 2023-01-13
Automate fastly ip addresses

cegerhardson:


pillar/base/firewall/fastly-logging.sls
```diff
@@ -4,12 +4,14 @@
 firewall:
 {% for address in fastly_ip_ranges.get('dict', {}).get('addresses', []) %}
   fastly_syslog_ipv4_{{ loop.index }}:
-    source: {{ address }}
+    source:
+      - {{ address }}
```
ewdurbin, line 8: The new rule for salt-lint is suggesting that nested dictionaries should be over indented. So in this case we should add two-spaces before `source`

pillar/base/firewall/fastly-logging.sls
```diff
@@ -4,12 +4,14 @@
...
     port: 514
 {% endfor %}
 
 {% for address in fastly_ip_ranges.get('dict', {}).get('ipv6_addresses', []) %}
   fastly_syslog_ipv6_{{ loop.index }}:
-    source6: {{ address }}
+    source6:
```
ewdurbin, line 14: The new rule for salt-lint is suggesting that nested dictionaries should be over indented. So in this case we should add two-spaces before `source6`

ewdurbin (changes requested):

ewdurbin (approved):

Issue #298, opened 2023-01-24
"module.run" is using its deprecated version and will expire in version "Phosphorus"


ewdurbin:
```
[WARNING ] The function "module.run" is using its deprecated version and will expire in version "Phosphorus".
```

Some of our states use `module.run` which apparently is using a deprecated syntax.

We should determine what the call should look like and update all invocations in our states to silence this warning.

To reproduce, `vagrant up` the salt-master and then `vagrant up` a node that uses the `module.run` function such as `cdn-logs`.

ewdurbin:
Little more investigation pairing with @cegerhardson, seems that this is only on the first invocation of state.highstate for some hosts. We also don't appear to using the `module.run` function directly. My best guess at this point is we're using it indirectly via a built-in salt state.

Attempted to get debug output by destroying cdn-logs vagrant box locally then bringing it back up with this patch applied:

```diff
diff --git a/Vagrantfile b/Vagrantfile
index 4763e96..671af6a 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -68,16 +68,16 @@ Vagrant.configure("2") do |config|
       apt-get install -y salt-minion
       echo 'master: #{MASTER1}\n' > /etc/salt/minion.d/local.conf
       service salt-minion restart
-      salt-call state.highstate
+      salt-call -l debug state.highstate
       while ! salt-call consul.cluster_ready | grep True; do echo 'waiting for consul'; sleep 1; done
       salt '*' saltutil.sync_all
       salt '*' saltutil.refresh_grains
       salt '*' saltutil.refresh_pillar wait=True timeout=30
-      salt-call state.highstate
+      salt-call -l debug state.highstate
     HEREDOC
 
     # Run this always, because we need to sync our states.
-    s_config.vm.provision :shell, inline: "salt-call state.highstate", run: "always"
+    s_config.vm.provision :shell, inline: "salt-call -l debug state.highstate", run: "always"
   end
 
   SERVERS.each_with_index do |server_c, num|
```

but didn't see debug output from vagrant up :/

cegerhardson:
After further investigation with @ewdurbin, it appears clear that the warning is coming from an indirect use via a built-in salt state, as we are not using `module.run` anywhere directly in our configuration. 

As an attempt to confirm this hypothesis, we updated salt to its latest configuration, which cleared out this warning. 
``` diff --git a/Dockerfile b/Dockerfile
index 1e214c8..efc696b 100644
--- a/Dockerfile
[...]

PR #301, opened 2023-01-27
pebble-source state failing

cegerhardson:
When bringing up the salt-master locally, the pebble-source state `git.latest` function fails with this comment: `Repository would be updated from 087582e to d5fa738, but there are uncommitted changes. Set 'force_reset' to True (or 'remote-changes') to force this update and discard these changes.`

By defualt, `git.lastest` `force_reset` is set to false. By changing our configuration to use `force_reset: remote-changes`, we can instruct[ salt](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.git.html#:~:text=in%20version%202015.8.0.-,force_resetFalse,-If%20the%20update) not to discard local changes if the repo is up-to-date with the remote repository.

### DIFF
--- a/salt/tls/pebble.sls
+++ b/salt/tls/pebble.sls
@@ -13,6 +13,7 @@ pebble-golang-workspace:
 pebble-source:
    git.latest:
      - name: https://github.com/letsencrypt/pebble.git
+     - force_reset: remote-changes
      - target: /usr/local/src/pebble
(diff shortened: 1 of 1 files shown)


cegerhardson:
seems like this change also addresses issue #298

ewdurbin (approved):

Issue #297, opened 2023-01-24
"Invalid compound target:  for results: " error when calling state.highstate


ewdurbin:
When calling `state.highstate` on any given node, the following log lines appear during execution:

```
[ERROR   ] Invalid compound target:  for results: 
[ERROR   ] Invalid compound target:  for results: 
```

We should determine what these are and how to resolve them.

Running `sudo salt-call -l debug state.highstate` after `vagrant ssh`ing into a given node locally should produce output that helps us track this down.

PR #302, opened 2023-02-07
Add a default option 

cegerhardson:
Adding a default option to prevent returning an empty string to `match.compound` removing error addressed in #297

Note: this locks up `devnull.psf.io` from being used in the future.

### DIFF
--- a/pillar/base/users.sls
+++ b/pillar/base/users.sls
@@ -248,6 +248,8 @@ users:
       mail:
         allowed: True
         sudo: True
+        groups:
+          - mailman
 
   hildeb:
     fullname: "Ralf Hildebrandt"
@@ -290,6 +292,8 @@ users:
       mail:
         allowed: True
         sudo: True
+        groups:
+          - mailman
 
   ambv:
     fullname: "Łukasz Langa"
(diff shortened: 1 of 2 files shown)


pillar/base/users.sls
```diff
@@ -251,8 +248,6 @@ users:
       mail:
         allowed: True
         sudo: True
-        groups:
-          - mailman
```
ewdurbin, line 255: We want to leave the `mailman` group for users on mail.python.org which also runs mailman. The `mailman` role was back when we had GNU mailman project infra hosted inside the PSF infra long ago.

pillar/base/users.sls
```diff
@@ -295,8 +290,6 @@ users:
       mail:
         allowed: True
         sudo: True
-        groups:
```
ewdurbin, line 298: We want to leave the `mailman` group for users on mail.python.org which also runs mailman. The `mailman` role was back when we had GNU mailman project infra hosted inside the PSF infra long ago.

ewdurbin (changes requested):
Once the mailman group bits are restored as noted above, this is good to go!

ewdurbin (approved):

Issue #299, opened 2023-01-24
`pkgrepo.managed` state reports Changed on every run


ewdurbin:
When running `state.highstate` on any of our hosts, the various repositories we add via the `pkgrepo.managed` state show as "Changed" on every run.

Perhaps we are incorrectly using this state or should move to another one. Ideally `state.highstate` reports no changes when run repeatedly.

PR #303, opened 2023-02-07
Removing trailing slash on `pkgrepo.managed` URLs 

cegerhardson:
Removing trailing slash on `pkgrepo.managed` URLs to prevent states from always reporting changed. This addresses issue #299 with reference to https://github.com/saltstack/salt/issues/58781


ewdurbin (approved):
Thanks @cegerhardson!

PR #305, opened 2023-02-21
Configure salt to manage psf_known_hosts and salt-server-list.rst

cegerhardson:
Configure salt to manage psf_known_hosts and salt-server-list.rst and serve them via HTTP

### DIFF
--- a/pillar/base/haproxy.sls
+++ b/pillar/base/haproxy.sls
@@ -94,6 +94,11 @@ haproxy:
       verify_host: salt.psf.io
       check: "GET /.well-known/acme-challenge/sentinel HTTP/1.1\\r\\nHost:\\ salt.psf.io"
 
+    publish-files:
+      domains: []
+      verify_host: salt-public.psf.io
+      check: "GET /srv/public HTTP/1.1\\r\\nHost:\\ salt-public.psf.io"
+
   redirects:
     cheeseshop.python.org:
       target: pypi.org
--- a/pillar/base/haproxy.sls
+++ b/pillar/base/haproxy.sls
@@ -95,9 +95,10 @@ haproxy:
       check: "GET /.well-known/acme-challenge/sentinel HTTP/1.1\\r\\nHost:\\ salt.psf.io"
 
     publish-files:
-      domains: []
-      verify_host: salt-public.psf.io
-      check: "GET /srv/public HTTP/1.1\\r\\nHost:\\ salt-public.psf.io"
+      domains:
+        - salt-public.psf.io
+      verify_host: salt.psf.io
+      check: "GET /salt-server-list.rst HTTP/1.1\\r\\nHost:\\ salt-public.psf.io"
 
   redirects:
     cheeseshop.python.org:
--- a/salt/base/config/publish-files-nginx.conf
+++ b/salt/base/config/publish-files-nginx.conf
@@ -1,0 +1,13 @@
+server {
+  listen 80 ssl default_server;
+
+  ssl_certificate /etc/ssl/private/salt.psf.io.pem;
+  ssl_certificate_key /etc/ssl/private/salt.psf.io.pem;
+
+  server_name _;
+
+  location / {
+      root /srv/public;
+      try_files $uri =404;
+  }
+}
--- a/salt/base/config/publish-files-nginx.conf
+++ b/salt/base/config/publish-files-nginx.conf
@@ -1,10 +1,10 @@
 server {
-  listen 9001 ssl salt-public.psf.io;
+  listen 9001 ssl;
 
   ssl_certificate /etc/ssl/private/salt.psf.io.pem;
   ssl_certificate_key /etc/ssl/private/salt.psf.io.pem;
 
-  salt-public.psf.io _;
+  server_name salt-public.psf.io;
 
   location / {
       root /srv/public;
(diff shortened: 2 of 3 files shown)


salt/base/salt.sls
```diff
@@ -80,21 +80,50 @@ salt-master:
     - require:
       - pkg: consul-pkgs
 
-/srv/psf_known_hosts:
+/srv/public:
+  file.directory:
+    - user: root
+    - group: root
+    - mode: "0644"
```
ewdurbin, line 87: we need "0755" here. Directory permissions 0644 -> `rw-r--r--` won't allow anyone to change directory in without execute, so 0755 -> `rwxr-xr-x` will let nginx read from the directory via try_files

salt/base/salt.sls
```diff
@@ -80,21 +80,50 @@ salt-master:
...
+    - require:
+       - file: /srv/public
+
+/etc/consul.d/service-publish-files.conf:
+  file.managed:
+    - source: salt://consul/etc/service.jinja
+    - template: jinja
+    - context:
+       name: publish-files
+       port: 9000
```
ewdurbin, line 119: This port conflicts with the service expoed by the `letsencrypt-well-known` service. We should bind to something else (9001 seems fine).

salt/base/config/publish-files-nginx.conf
```diff
@@ -0,0 +1,13 @@
+server {
+  listen 80 ssl default_server;
```
ewdurbin, line 2: This needs to match the port we set in the service config (9001 was suggested).

salt/base/config/publish-files-nginx.conf
```diff
@@ -0,0 +1,13 @@
...
+
+  ssl_certificate /etc/ssl/private/salt.psf.io.pem;
+  ssl_certificate_key /etc/ssl/private/salt.psf.io.pem;
+
+  server_name _;
```
ewdurbin, line 7: Let's choose a hostname, which will allow us to remove `default_server` above and set `server_name` to something precise.

Let's go with `salt.psf.io`.

ewdurbin (changes requested):
Some changes suggested to get things working.

Otherwise we'll also need to expose the new `publish-files` service via the loadbalancer.

See notes in https://infra.psf.io/services/discovery/ on how to expose a service registered with consul via haproxy.

salt/base/config/publish-files-nginx.conf
ewdurbin: perhaps `salt-public.psf.io` to keep from confusing the internal name we already use (salt.psf.io).

salt/base/salt.sls
```diff
@@ -80,21 +80,50 @@ salt-master:
...
     - group: root
     - mode: "0644"
+
+/etc/nginx/sites.d/publish-files.conf:
+  file.managed:
+    - source: salt://base/config/publish-files-nginx.conf
+    - user: root
+    - group: root
+    - require:
+       - file: /srv/public
```
ewdurbin, line 111: we probably want this to require that sites.d exists too, in case we ever reorder/remove the letsencrypt-well-known service which executes before this.

```suggestion
       - file: /etc/nginx/sites.d/
       - file: /srv/public
```

salt/base/salt.sls
```diff
@@ -80,21 +80,50 @@ salt-master:
...
     - template: jinja
     - user: root
     - group: root
     - mode: "0644"
+
+/etc/nginx/sites.d/publish-files.conf:
+  file.managed:
+    - source: salt://base/config/publish-files-nginx.conf
+    - user: root
+    - group: root
```
ewdurbin, line 109: always good to set explicit file permissions on configuration files.

```suggestion
    - group: root
    - mode: "0644"
```

pillar/base/haproxy.sls
```diff
@@ -94,6 +94,11 @@ haproxy:
       verify_host: salt.psf.io
       check: "GET /.well-known/acme-challenge/sentinel HTTP/1.1\\r\\nHost:\\ salt.psf.io"
 
+    publish-files:
+      domains: []
```
ewdurbin, line 98: this should be set to a list containing the domain we are serving from (`salt-public.psf.io`), otherwise requests won't be routed to it. the empty domains block above is for lets-encrypt which needs to respond to any domain.

pillar/base/haproxy.sls
```diff
@@ -94,6 +94,11 @@ haproxy:
...
+      verify_host: salt-public.psf.io
```
ewdurbin, line 99: verify_host is for the backend hostname we'll verify the certificate against, so in this case it needs to match the certificate issued to the salt-master: `salt.psf.io`.

pillar/base/haproxy.sls
```diff
@@ -94,6 +94,11 @@ haproxy:
...
+      check: "GET /srv/public HTTP/1.1\\r\\nHost:\\ salt-public.psf.io"
```
ewdurbin, line 100: I'm not sure if this health check will pass based on how [try_files](https://nginx.org/en/docs/http/ngx_http_core_module.html#try_files) is configured.

ewdurbin: ah, yeah and the path is extraneous since nginx is configured to serve from /srv/public.

We could just have it `GET /salt-server-list.rst` for a check.

ewdurbin (approved):

ewdurbin:
Looks great @cegerhardson!

ewdurbin:
I tried rolling this out but it doesn't seem to be working.

First issue is it doesn't appear that the service is being registered with consul. Could you take a look?

ewdurbin:
And as a sanity check, I tried to access the new server from the load balancer, but it doesn't seem port 9001 is accessible from other hosts.

PR #308, opened 2023-02-24
Turn on firewall

cegerhardson:


ewdurbin:
Disabled in https://github.com/python/psf-salt/pull/245/commits/eae522b216e905e62d9e134fc13e968ba697b5ad for unknown reasons. I seem to recall iptables _not_ working... but this was just a confusion waiting to happen. Thanks @cegerhardson

PR #312, opened 2023-03-23
Fix /ftp redirect on docs.python.org for archives

encukou:
The `^~` modifier makes this rule take priority priority over regex
(specifically the `pdf|zip|epub|bz2` regex above), see
https://nginx.org/en/docs/http/ngx_http_core_module.html#location

See: https://github.com/python/cpython/issues/95290


encukou:
I haven't tested this -- I don't have the setup for that.

PR #314, opened 2023-05-22
Update users.sls

msapiro:
Add msapiro and maxking to the mailman group on gnumailman.


ewdurbin (approved):

ewdurbin:
@msapiro should be live in 11 minutes

msapiro:
Thanks Ee.

PR #317, opened 2023-08-22
Rate limit

cegerhardson:
Configure bugs server to allow for nginx rate-limiting.

### DIFF
--- a/salt/bugs/config/nginx.conf.jinja
+++ b/salt/bugs/config/nginx.conf.jinja
@@ -3,7 +3,7 @@ log_format timed_combined_{{ tracker }} '$remote_addr - $remote_user [$time_loca
     '"$http_referer" "$http_user_agent" '
     '$request_time $upstream_response_time $pipe';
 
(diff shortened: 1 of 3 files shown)


salt/bugs/config/nginx.conf.jinja
```diff
@@ -3,6 +3,9 @@ log_format timed_combined_{{ tracker }} '$remote_addr - $remote_user [$time_loca
     '"$http_referer" "$http_user_agent" '
     '$request_time $upstream_response_time $pipe';
 
+limit_req_zone $remote_addr zone=mylimit-{{ tracker }}:10m rate=1r/s;
```
ewdurbin, line 6:
```suggestion
limit_req_zone $binary_remote_addr zone=limit-{{ tracker }}:10m rate=1r/s;
```

Change to binary_remote_addr rather than remote_addr, per note in https://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req_zone

Also remove the tutorial name "mylimit" in favor of just limit.

ewdurbin (approved):

PR #318, opened 2023-09-14
Rate limit hg

cegerhardson:
Enable apache mod_ratelimit module to limit bandwidth of the client.

### DIFF
--- a/salt/hg/config/hg.apache.conf.jinja
+++ b/salt/hg/config/hg.apache.conf.jinja
@@ -1,6 +1,5 @@
 <IfModule mod_qos.c>
-    QS_ClientEventLimitCount 20 60
-    QS_ClientEventBlockCount 20 300
+    QS_ClientEventLimitCount 100 60
     QS_ErrorResponseCode 429
 </IfModule>
 
--- a/salt/hg/config/hg.apache.conf.jinja
+++ b/salt/hg/config/hg.apache.conf.jinja
@@ -1,4 +1,11 @@
+#Configure rate limiting
+QS_ClientEventLimitCount 10 60 rate_limit
+QS_ClientEventPerSecLimit 1
+QS_ErrorResponseCode 429
+
 <VirtualHost _default_:9000>
+    SetEnvIf Request_URI / rate_limit=1
+
     ServerName hg.psf.io
     ServerAlias hg.python.org
 
--- a/salt/hg/config/hg.apache.conf.jinja
+++ b/salt/hg/config/hg.apache.conf.jinja
@@ -3,6 +3,14 @@ QS_ClientEventLimitCount {{pillar.get("rate_limit_config", {}).get("requests", 1
 QS_ClientEventPerSecLimit 1
 QS_ErrorResponseCode {{pillar.get("rate_limit_config", {}).get("response_code", 429)}}
 
+QS_VipIpHeaderName My-Header drop
+RemoteIPHeader X-Forwarded-For
+RemoteIPInternalProxy {{pillar["psf_internal_network"] }}
+
+<If "%{REMOTE_ADDR} -ipmatch '{{ pillar["psf_internal_network"] }}'">
+ Header set My-Header "inside"
+</If>
+
 <VirtualHost _default_:9000>
     SetEnvIf Request_URI / rate_limit=1
 
--- a/salt/hg/config/hg.apache.conf.jinja
+++ b/salt/hg/config/hg.apache.conf.jinja
@@ -1,10 +1,6 @@
-<IfModule mod_qos.c>
-    QS_ClientEventLimitCount 100 60
-    QS_ErrorResponseCode 429
-</IfModule>
-
 <VirtualHost _default_:9000>
-    ServerName hg.python.org
+    ServerName hg.psf.io
+    ServerAlias hg.python.org
 
     SSLEngine on
     SSLCertificateFile /etc/ssl/private/hg.psf.io.pem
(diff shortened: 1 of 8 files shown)


salt/hg/config/legacy.apache.conf.jinja
```diff
@@ -116,4 +116,12 @@
     <Location /dev/peps>
         Header set X-Robots-Tag "noindex, nofollow"
     </Location>
+
+  <IfModule mod_ratelimit.c>
+    <Location /srv/hg/>
```
ewdurbin, line 121: I think the mercurial servers configured is in https://github.com/python/psf-salt/blob/main/salt/hg/config/hg.apache.conf.jinja

Its unclear if this configuration here will affect that server.

salt/hg/init.sls
```diff
@@ -234,6 +234,8 @@ apache2:
     - mode: "0644"
     - require:
       - pkg: apache2
+    - cmd.run:
+      - name: a2enmod ratelimit
```
ewdurbin, line 238: I don't think we want to run this every time salt runs... maybe we could use the apache_module state just once? https://docs.saltproject.io/en/latest/ref/states/all/salt.states.apache_module.html

salt/hg/init.sls
```diff
@@ -287,6 +289,8 @@ apache2:
     - mode: "0644"
     - require:
       - file: /etc/apache2/legacy-redirects.conf
+    - cmd.run:
```
ewdurbin, line 292: This seems duplicated.

salt/hg/config/svn.apache.conf.jinja
```diff
@@ -44,6 +44,13 @@
         Require valid-user
     </Location>
 
+  <IfModule mod_ratelimit.c>
+    <Location /srv/hg/>
```
ewdurbin, line 48: Similarly to the other config, I'm not sure that the hg server would be impacted by this change, just the svn server.

ewdurbin (commented):
Overall, I'm not positive that rate limiting via bandwidth is what we want, but rather by request rate.

Is that something supported?

salt/hg/config/hg.apache.conf.jinja
cegerhardson, line 1: With reference to `mod_qos`: 

- https://mod-qos.sourceforge.net/index.html#QS_ClientEventBlockCount
- https://mod-qos.sourceforge.net/index.html#QS_ClientEventLimitCount

salt/hg/init.sls
```diff
@@ -152,6 +153,10 @@ apache2:
       - file: /etc/apache2/sites-enabled/*
       - file: /etc/apache2/mods-enabled/*
       - file: /etc/ssl/private/hg.psf.io.pem
+  cmd.run:
```
ewdurbin, line 156: Still think we could use the apache module state instead of a command! https://docs.saltproject.io/en/latest/ref/states/all/salt.states.apache_module.html

salt/hg/config/hg.apache.conf.jinja
```diff
@@ -1,3 +1,9 @@
+<IfModule mod_qos.c>
+    QS_ClientEventLimitCount 20 60
+    QS_ClientEventBlockCount 20 300
```
ewdurbin, line 3: I don't think we need to be blocking quite yet. Let's just configure a limit.

salt/hg/config/hg.apache.conf.jinja
```diff
@@ -1,3 +1,9 @@
+<IfModule mod_qos.c>
```
ewdurbin, line 1: I guess we just need to make sure `libapache2-mod-qos` is installed from apt and enabled via a state (as referenced in other comment).

ewdurbin: Ope! missed that it was! This is resolved.

salt/hg/config/hg.apache.conf.jinja
```diff
@@ -1,5 +1,13 @@
+#Configure rate limiting
+QS_ClientEventLimitCount 10 60 rate_limit
+QS_ClientEventPerSecLimit 1
+QS_ErrorResponseCode 429
+
 <VirtualHost _default_:9000>
-    ServerName hg.python.org
+    SetEnvIf Request_URI / rate_limit=1
+
+    ServerName hg.psf.io
```
ewdurbin, line 9: I don't think we should ship this change to production. Please refer the ServerName/ServerAlias parts.

salt/hg/config/hg.apache.conf.jinja
```diff
@@ -1,5 +1,13 @@
+#Configure rate limiting
+QS_ClientEventLimitCount 10 60 rate_limit
```
ewdurbin, line 2: This looks great! I think the one suggestion I would have is that the "default" global rate limit be more permissive, maybe 120 per 60 seconds. We can then configure `QS_ClientEventPerSecLimit` inside a given `VirtualHost`

ewdurbin (changes requested):
Great progress @cegerhardson!

I think as a next step we should extract the global configuration (`QS_ClientEventLimitCount`, `QS_ClientEventPerSecLimit`, and `QS_ErrorResponseCode`) to its own top-level configuration (similar to ports.conf).

Then all we should need to do is set a rate limit with `QS_ClientEventPerSecLimit` inside each of our services on this host (hg, svn, and legacy).

salt/hg/config/hg.apache.conf.jinja
cegerhardson: this makes total sense

ewdurbin (commented):
Great progress! This looks like it should be rate limiting!

As discussed next steps are to ensure the true client IP is making it to Apache, and to ensure we never rate limit our own internal ip addresses!

PR #324, opened 2023-12-13
mod_remoteip for X-Forwarded-For handling 

cegerhardson:
This PR introduces a globally applied `mod_remoteip` configuration for handling client IPs, preparing for future rate-limiting configurations. It also specifies a salt state to manage `remoteip.conf`, creates a symlink to define the configuration, and another symlink to enable the configuration.

To verify these changes locally: 
1. bring up the salt-master, `laptop:psf-salt user$ vagrant up salt-master`
2. bring up the loadbalancer, `laptop:psf-salt user$ vagrant up loadbalancer`
3. bring up hg, `laptop:psf-salt user$ vagrant up hg`
4. in another window, ssh into the vagrant hg box, `laptop:psf-salt user$  vagrant ssh hg`
5. in another window, from your local computer, run the curl command `curl -k https://127.0.0.1:20010/hello -H "Host: hg.python.org"`
6. in the vagrant box, run `sudo tail -f /var/log/apache2/hg.access.log | grep hello`

### DIFF
--- a/salt/hg/init.sls
+++ b/salt/hg/init.sls
@@ -171,6 +171,10 @@ apache2:
       - file: /etc/apache2/mods-enabled/*
       - file: /etc/ssl/private/hg.psf.io.pem
 
+/etc/apache2/mods-enabled/remoteip.load:
+  file.symlink:
+    - target: /etc/apache2/mods-available/remoteip.load
+
 /etc/apache2/mods-enabled/headers.load:
   file.symlink:
     - target: /etc/apache2/mods-available/headers.load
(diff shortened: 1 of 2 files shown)


salt/hg/init.sls
```diff
@@ -171,6 +171,10 @@ apache2:
       - file: /etc/apache2/mods-enabled/*
```
ewdurbin, line 171: We probably want to add `/etc/apache2/conf-enabled/*` to this watch so that configuration changes in that directory reloads apache automatically!

ewdurbin (approved):
This is working as expected! I think its good to go after that small change to ensure future updates to conf-enabled/available result in a server reload.

salt/hg/init.sls
cegerhardson: Roger that 🫡

PR #327, opened 2023-12-19
Begin logging X-Forwarded-For in haproxy frontends

ewdurbin:
This will allow for more indepth debugging of what's going on with the DigitalOcean loadbalancer and how it is interacting with haproxy's X-Forwarded-For handling.

Currently hg.python.org _should_ be seeing such headers and acting accordingly thanks to https://github.com/python/psf-salt/pull/324, but as is it is not working :(


cegerhardson:
Looks like this is going to be very helpful, thanks Ee :-)

PR #334, opened 2024-01-16
guard older distros from jammy aptkey support

ewdurbin:
older versions of Ubuntu that we currently support/run do not play nicely with the changes added to address apt-key deprecation.

this guards the new changes and applies them only to jammy boxes

### DIFF
--- a/salt/base/repo.sls
+++ b/salt/base/repo.sls
@@ -1,16 +1,13 @@
-/etc/apt/keys:
-  file.directory:
-    - user: root
-    - group: root
-    - dir_mode: "0755"
-    - file_mode: "0644"
-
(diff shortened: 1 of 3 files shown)


ewdurbin:
I was unable to bring up non-jammy boxes before making these changes. @cegerhardson

cegerhardson:
Ah, yes. A slight oversight not considering this deprecation is only for 22.04, makes sense it wasn't working outside of jammy.

cegerhardson (approved):
Ah, yes. A slight oversight not considering this deprecation is only for 22.04, makes sense it wasn't working outside of jammy. Good catch.

PR #261, opened 2022-07-25
Upgrade codespeed configuration to run in development

cegerhardson:
The current status of this PR is that all states are applying aside from:

``` codespeed:           ID: codespeed-cpython-pre-reload
    codespeed:     Function: cmd.run
    codespeed:         Name: /srv/codespeed/cpython/env/bin/python manage.py migrate --noinput && /srv/codespeed/cpython/env/bin/python manage.py collectstatic --noinput
    codespeed:       Result: False
    codespeed:      Comment: Command "/srv/codespeed/cpython/env/bin/python manage.py migrate --noinput && /srv/codespeed/cpython/env/bin/python manage.py collectstatic --noinput" run
    codespeed:      Started: 15:27:47.171804
    codespeed:     Duration: 260.412 ms
    codespeed:      Changes:
    codespeed:               ----------
    codespeed:               pid:
    codespeed:                   1398788
    codespeed:               retcode:
    codespeed:                   1
    codespeed:               stderr:
    codespeed:                   Traceback (most recent call last):
    codespeed:                     File "manage.py", line 9, in <module>
    codespeed:                       execute_from_command_line(sys.argv)
    codespeed:                     File "/srv/codespeed/cpython/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    codespeed:                       utility.execute()
    codespeed:                     File "/srv/codespeed/cpython/env/lib/python3.8/site-packages/django/core/management/__init__.py", line 328, in execute
    codespeed:                       django.setup()
    codespeed:                     File "/srv/codespeed/cpython/env/lib/python3.8/site-packages/django/__init__.py", line 18, in setup
    codespeed:                       apps.populate(settings.INSTALLED_APPS)
    codespeed:                     File "/srv/codespeed/cpython/env/lib/python3.8/site-packages/django/apps/registry.py", line 108, in populate
    codespeed:                       app_config.import_models(all_models)
    codespeed:                     File "/srv/codespeed/cpython/env/lib/python3.8/site-packages/django/apps/config.py", line 198, in import_models
    codespeed:                       self.models_module = import_module(models_module_name)
    codespeed:                     File "/usr/lib/python3.8/importlib/__init__.py", line 127, in import_module
    codespeed:                       return _bootstrap._gcd_import(name[level:], package, level)
    codespeed:                     File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
[...]

PR #335, opened 2024-01-16
restore codespeed local development

ewdurbin:
This allows users to get a (almost?) working codespeed instance up locally:

```
vagrant up salt-master
vagrant up loadbalancer
vagrant up codespeed
```

Currently when isolated, cpython codespeed even works!

```shell
$ curl -sk https://localhost:20010/ -H"Host: speed.python.org" | head

<!DOCTYPE html>
<html>
  <head>
    <title>
	Python Speed Center
</title>
    <meta name="description" content="A performance analysis tool for software projects. It shows performance regresions and allows comparing different applications or implementations">
    <meta name="keywords" content="performance, test, plots, charts">
    <meta charset="UTF-8">
```


ewdurbin:
d'oh, one note: I did have to manually install `psycopg2-binary` into the cpython speed virtual environment in order to make the above work, but that is outside scope of the infra and should be handled in https://github.com/python/codespeed

ewdurbin:
cc @ambv

salt/codespeed/config/django-settings.py.jinja
```diff
@@ -14,7 +14,7 @@ DATABASES = {
         "USER": "{{ db_user}}",
         "PASSWORD": "{{ db_pass }}",
         "OPTIONS": {
-            "sslmode": "verify-full",
+            "sslmode": "verify-ca",
```
ewdurbin, line 17: this is a local kludge and should be handled differently. for simplicity... I think we should probably just have the salt-master run a a postgres cluster for local dev so we don't have to fire up yet another box.

ewdurbin:
supersedes: #261

PR #350, opened 2024-02-10
Update users.sls to allow my return to the wiki

smontanaro:


ewdurbin (approved):

ewdurbin:
This will be live in 12 minutes on the hour (next run of the */15 cron job that distributes these changes).

PR #351, opened 2024-02-15
Add user for Petr Viktorin with buildbot access

encukou:
Hello,
I triage Buildbot results as part of my Deputy Developer in Residence job. I'd appreciate having access so I can apply config changes effectively.


ewdurbin (approved):

ewdurbin:
should be live shortly! automation runs on */15 cron so shortly after :00

PR #355, opened 2024-05-09
Starting my migration out of RSA 2048.

JulienPalard:
You can « verify » it's legit by running:

    ssh docs.nyc1.psf.io cat /home/psf-users/mdk/migrating-out-of-rsa-2048


ewdurbin:
```
ee@docs:~$ sudo cat /home/psf-users/mdk/migrating-out-of-rsa-2048
Yes PR https://github.com/python/psf-salt/pull/355 is from me.
```

PR #356, opened 2024-05-20
Add downloads/docs access for the release manager of Python 3.14 and 3.15 (3.26?)

ambv:
See: https://peps.python.org/pep-0745/


ewdurbin:
Cross validated with https://github.com/hugovk.keys

PR #365, opened 2024-07-02
add cursory support for Ubuntu 24.04 "noble"

ewdurbin:
salt-master does come up clean with this, but haven't verified further


JacobCoffee:
https://github.com/python/psf-salt/pull/367#issuecomment-2229474552

PR #372, opened 2024-07-10
Update buildbot nginx config to do away with `^/all/`

zware:
I'm not confident in all of these changes, but hopefully it's enough to get my point across :)


salt/buildbot/config/nginx.conf.jinja
```diff
@@ -21,18 +21,13 @@ server {
...
-  rewrite ^/stable(/?)$ /all/#/grid?tag=stable redirect;
+  rewrite ^/3.(\d+)(/?)$ /#/grid?branch=3.$1 redirect;
+  rewrite ^/3.x(/?)$ /#/grid?branch=main redirect;
+  rewrite ^/3.(\d+).stable(/?)$ /#/grid?branch=3.$1&tag=stable redirect;
+  rewrite ^/3.x.stable(/?)$ /#/grid?branch=main&tag=stable redirect;
+  rewrite ^/stable(/?)$ /#/grid?tag=stable redirect;
+
+  rewrite ^/all/(.*)$ /$1 redirect;
 
   location @backend {
```
zware, line 32: Actually, I think this might be better now:

```suggestion
  location / {
```

(along with removing the other `location /` section below)

JacobCoffee: that works i think!

zware: Done.

JacobCoffee (approved):
lets try it! 😬 I havent had any bad things come up in my error.log yet.

PR #376, opened 2024-07-15
buildbot: Move /all rewrite out of location /

zware:
This allows wss://buildbot.python.org/all/ws to load again.


ewdurbin:
@zware, applied.

PR #384, opened 2024-07-16
feat(base): ensure only one apt source for salt repo

JacobCoffee:
## What

Currently we get two things in `/etc/apt/source.list.d/`:
- `salt.list`
- `saltstack.list`
this is not fun

make it fun, squish into one


ewdurbin:
Perhaps we might want to add a salt state temporarily to ensure the old file is removed?

ewdurbin (approved):

ewdurbin (approved):

PR #386, opened 2024-07-16
fix(base): disable ntp installed on noble

JacobCoffee:
## What

- Does not install ntp on noble as it uses timesyncd as an ntp client and running highstate in production causes conflicts

### DIFF
--- a/salt/base/sanity.sls
+++ b/salt/base/sanity.sls
@@ -5,6 +5,7 @@ niceties:
       - htop
       - traceroute
 
+{% if grains["oscodename"] != "noble" %}
 time-sync:
   pkg.installed:
     - pkgs:
@@ -15,7 +16,7 @@ ntp:
   service:
     - running
     - enable: True
-
+{% endif %}
 
 # Cron has a default $PATH of only /usr/bin:/bin, however the root user's
 # default $PATH in the shell includes various sbin directories. This can cause
--- a/salt/base/sanity.sls
+++ b/salt/base/sanity.sls
@@ -16,7 +16,7 @@ systemd-timesyncd:
     - running
     - enable: True
 {% else %}
-time-sync-ntp
+time-sync-ntp:
   pkg.installed:
     - pkgs:
       - ntp
--- a/salt/base/sanity.sls
+++ b/salt/base/sanity.sls
@@ -10,8 +10,12 @@ systemd-timesyncd:
    pkg:
      - installed
    service:
+     {% if grains["detect_virt"] in ["docker"] %}
+     - enabled
+     {% else %}
      - running
      - enable: True
+    {% endif %}
 {% else %}
 ntp-packages:
   pkg.installed:
--- a/salt/base/sanity.sls
+++ b/salt/base/sanity.sls
@@ -5,8 +5,18 @@ niceties:
       - htop
       - traceroute
 
-{% if grains["oscodename"] != "noble" %}
-time-sync:
+{% if grains["oscodename"] in ["noble"] %}
+time-sync-timesyncd:
+    pkg.installed:
+      - pkgs:
+        - systemd-timesyncd
+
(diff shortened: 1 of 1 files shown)


JacobCoffee:
Confusion on why this doesnt affect local

salt/base/sanity.sls
```diff
@@ -5,6 +5,7 @@ niceties:
       - htop
       - traceroute
 
+{% if grains["oscodename"] != "noble" %}
```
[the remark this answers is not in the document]
JacobCoffee: see [da270e1](https://github.com/python/psf-salt/pull/386/commits/da270e110440cc3c6c363c4797f8e10ecd2e420d)

salt/base/sanity.sls
```diff
@@ -5,7 +5,18 @@ niceties:
...
+    pkg.installed:
+      - pkgs:
+        - systemd-timesyncd
+
+systemd-timesyncd:
+  service:
+    - running
+    - enable: True
+{% else %}
+time-sync-ntp
```
JacobCoffee, line 19:
```suggestion
time-sync-ntp:
```

salt/base/sanity.sls
```diff
@@ -5,7 +5,18 @@ niceties:
...
       - traceroute
 
-time-sync:
+{% if grains["oscodename"] in ["noble"] %}
+time-sync-timesyncd:
+    pkg.installed:
+      - pkgs:
+        - systemd-timesyncd
+
+systemd-timesyncd:
```
ewdurbin, line 14: This and the above block could be simplified a little to collapse it all into one top level key ala:

```
systemd-timesyncd:
  pkg:
    - installed
  service:
    - enabled: True
    - running
```

ewdurbin (commented):
Further down the rabbit hole, it seems systemd-timesyncd _refuses_ to run in a container, which is actually kinda cool but shows the limits of our local dev setup 🙃 

I opened #387 which should give us a way to bail out of functionality that doesn't work in a docker container via a grain, might be helpful in pushing this along.

salt/base/sanity.sls
ewdurbin: Note: this kinda stuff only works in situations where the names all line up, in this case the service name and package name are identical. To see this taken to an extreme, our nginx state uses it to great effect. https://github.com/python/psf-salt/blob/f7803fd0142862bb3289b5944c598801f290fefd/salt/nginx/init.sls#L8-L53

ewdurbin: [ad57881](https://github.com/python/psf-salt/pull/386/commits/ad5788177e30015fd0326fe98f451cd8139b2af0) looks great!

salt/base/sanity.sls
```diff
@@ -5,7 +5,15 @@ niceties:
       - htop
       - traceroute
 
-time-sync:
+{% if grains["oscodename"] in ["noble"] %}
+systemd-timesyncd:
+   pkg:
+     - installed
+   service:
+     - running
```
ewdurbin, line 13: now that we have `grains['detect_virt']` should we guard the `running` with that to make this not error locally?

I think in order to do that you'll have to change the syntax for the docker case to be `- enabled` rather than `- enable: True` (see [salt docs](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.service.html#salt.states.service.enabled)).

In total I think the result would be:

```
  service:
    {% if grains["detect_virt"] in ["docker"] %}
    - enabled
    {% else %}
    - running
    - enable: True
    {% endif %}
```

ewdurbin: main issue is since this service _will not_ start in docker the salt state for `running` will fail.

JacobCoffee: doh

ewdurbin (approved):

PR #393, opened 2024-07-17
feat(bugs): move service behind load balancers

JacobCoffee:
## What

- Moves the `bugs.{python,jython}.org`,  `issues.roundup-tracker.org` services behind the load balancers


pillar/base/firewall/bugs.sls
```diff
@@ -9,3 +9,6 @@ firewall:
     port: 587
   submission:
     port: 465
+  frontend-planet:
```
JacobCoffee, line 12:
```suggestion
  frontend-bugs:
```

pillar/base/haproxy.sls
```diff
@@ -67,6 +67,20 @@ haproxy:
       verify_host: planet.psf.io
       check: "HEAD / HTTP/1.1\\r\\nHost:\\ planet.psf.io"
 
+    bugs:
```
JacobCoffee, line 70: Need to verify this

salt/bugs/config/nginx.conf.jinja
```diff
@@ -77,3 +77,49 @@ server {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
   }
 }
+
+server {
```
JacobCoffee, line 81: also this

JacobCoffee (commented):
also need to understand `pillar/dev/secrets/tls/certs/loadbalancer.sls` generation

ewdurbin:
> also need to understand `pillar/dev/secrets/tls/certs/loadbalancer.sls` generation

this is just a stand-in for the production secrets (which live on the salt host itself and are cloned to /srv/pillar/prod/secrets).

so its not generated as much as the self-signed certs are copy/pasted from another block when adding a new domain. this is just so a valid (self-signed so not verifiable) cert is available for haproxy to start with in dev.

you can copy any stanza you want for bugs (but a secret is they're all the same 😂)

ewdurbin:
Hrmph. I'm concerned that the postfix setup for issues.roundup-tracker.org will be a bit of a headache for moving this behind the load balancer.

The main issue is that we'd need to poke holes for the mail services (25, 587, 465) running on issues.roundup-tracker.org through haproxy (similar to what we do for hg and buildbot [here)](https://github.com/python/psf-salt/blob/7df453e99fbdca1c762608dc0e108e107960a1e0/pillar/base/haproxy.sls#L153-L170). I think buildbot is the only one of these doing something sane, and as a matter of fact think we could probably stop serving 22 for hg at this point.

Our hiccups are that since the mail services are protocol defined ports...

1) We can only run one of them without some kind of SNI type host support to route traffic
2) Spammers/scanners going for any service fronted by the load balancer will probably all land in that postfix

I think we should consider this a bit more before proceeding... Its far from ideal :(

ewdurbin:
```diff
diff --git a/pillar/base/haproxy.sls b/pillar/base/haproxy.sls
index c3c0096..694b666 100644
--- a/pillar/base/haproxy.sls
+++ b/pillar/base/haproxy.sls
@@ -173,6 +173,20 @@ haproxy:
       bind: :20100
       service: hg-ssh
 
+{# Binding ports on haproxy boxes that can be pointed to by DigitalOcean load balancer #}
+    roundup_smtp:
+      bind: :20025
+      service: roundup-smtp
+
+    roundup_smtp:
+      bind: :20587
+      service: roundup-ssmtp
+
+    roundup_submission:
+      bind: :20465
+      service: roundup-submission
+{# end ports for DO loadbalancer #}
+
     buildbot_worker:
       bind: :20101
       service: buildbot-master-worker
diff --git a/salt/bugs/init.sls b/salt/bugs/init.sls
index 05bbe9a..da39c00 100644
--- a/salt/bugs/init.sls
+++ b/salt/bugs/init.sls
@@ -210,6 +210,21 @@ postfix:
       - file: /etc/postfix/virtual
       - file: /etc/postfix/reject_recipients
 
+{% for (port, service) in [(25, "smtp"), (587, "sstmp"), (465, "submission")] %}
+/etc/consul.d/service-roundup-{{ service }}.json:
+  file.managed:
+    - source: salt://consul/etc/service.jinja
+    - template: jinja
+    - context:
+        name: roundup-{{ service }}
+        port: {{ port }}
+    - user: root
+    - group: root
+    - mode: "0644"
+    - require:
+      - pkg: consul-pkgs
+{% endfor %}
+
 {% for tracker, config in pillar["bugs"]["trackers"].items() %}
 tracker-{{ tracker }}-database:
```

Here's a basic sketch of how we might bind the ports on the haproxy load balancers.

We could then setup a DigitalOcean load balancer to map the public protocol port to the specific port on our haproxy load balancers Ala:

- 80 -> 80
- 443 -> 443
- 25 -> 20025
- 587 -> 20587
- 465 -> 20465

salt/bugs/config/nginx.conf.jinja
JacobCoffee: these should be split

pillar/base/haproxy.sls
```diff
@@ -67,6 +67,29 @@ haproxy:
       verify_host: planet.psf.io
       check: "HEAD / HTTP/1.1\\r\\nHost:\\ planet.psf.io"
 
+    {% for (port, service) in [(25, "smtp"), (587, "smtps"), (465, "submission")] %}
```
ewdurbin, line 70: I think this needs to be in the salt state for `bugs`, not the haproxy configuration file.

probably around [here](https://github.com/python/psf-salt/blob/448b8a7c0007bf3ffaf7098b3363956401e82929/salt/bugs/init.sls#L212).

salt/haproxy/config/haproxy.cfg.jinja
```diff
@@ -87,9 +87,10 @@ listen tls:
...
+    bind 0.0.0.0:443 ssl alpn h2,http/1.1 crt star.python.org.pem crt star.pypa.io.pem crt star.pyfound.org.pem crt speed.pypy.org.pem crt www.pycon.org.pem crt jython.org.pem crt pypy.org.pem crt salt-public.psf.io.pem crt planetpython.org.pem crt bugs.python.org.pem crt bugs.jython.org.pem crt issues.roundup-tracker.org.pem crt mail.roundup-tracker.org.pem
+    bind :::443 ssl alpn h2,http/1.1 crt star.python.org.pem crt star.pypa.io.pem crt star.pyfound.org.pem  crt speed.pypy.org.pem crt www.pycon.org.pem crt jython.org.pem crt pypy.org.pem crt salt-public.psf.io.pem crt planetpython.org.pem crt bugs.python.org.pem crt bugs.jython.org.pem crt issues.roundup-tracker.org.pem crt mail.roundup-tracker.org.pem
+    bind :20010 ssl alpn h2,http/1.1 crt star.python.org.pem crt star.pypa.io.pem crt star.pyfound.org.pem crt speed.pypy.org.pem crt www.pycon.org.pem crt jython.org.pem crt pypy.org.pem crt salt-public.psf.io.pem crt planetpython.org.pem crt bugs.python.org.pem crt bugs.jython.org.pem crt issues.roundup-tracker.org.pem crt mail.roundup-tracker.org.pem
```
ewdurbin, line 92: for these and all lines above, we'll condense the necessary certs all into one by just adding [SANs](https://en.wikipedia.org/wiki/Subject_Alternative_Name#:~:text=Subject%20Alternative%20Name%20(SAN)%20is,Subject%20Alternative%20Names%20(SANs).), so we just need to add `crt bugs.python.org`.

pillar/dev/secrets/tls/certs/loadbalancer.sls
```diff
@@ -469,3 +469,159 @@ tls:
...
+      s8RvMB8GA1UdIwQYMBaAFPJrXEC964Djv1KtiYGjRFpDs8RvMA8GA1UdEwEB/wQF
+      MAMBAf8wDQYJKoZIhvcNAQELBQADggEBAGlJ+N5txBsBekRMkl2pGxUecihJWLXM
+      pwnXuhKswrsCpLiJlWijTWVBULfVn71rEfnMFNgdVn4i1TddgyK4cViHWZPBYcGd
+      SYbQK40xmLuIAJKM8uARdm99AmavKCH+ha6jFY8fZoU0+m51hOztXfGTIkLpLr2r
+      +0ydepkbAWqNH6NYNpUQKFxSlyTYvwaHUh0YzXMxgOj+foJCygyVnB/E7Fja92Ho
+      Pe93No9Ze0Jou4GsXmP2E1YY0i3jkCigmuVTQSrl85uxxHfHWNgr9OwN8ASoF9dp
+      ogsOBi74M0k7Ihp96JK6lUXTY+WnlJ3C9FZdByeXq6O4HLhgq5jug7E=
+      -----END CERTIFICATE-----
+
+    issues.roundup-tracker.org: |
```
ewdurbin, line 525: see note on Subject Alternative Names (SANs) in other file. we'll only need to add one entry here.

salt/haproxy/config/haproxy.cfg.jinja
JacobCoffee: do we need to proxy smtp here?

JacobCoffee: SANS set in https://github.com/python/psf-salt/blob/3a4c7693bb551a3180071b4517d57a7fd9c7a741/pillar/base/bugs.sls#L2-L6 right?

ewdurbin: yep!

ewdurbin: > do we need to proxy smtp here?

We would set the proxying up using a configuration like [these](https://github.com/python/psf-salt/blob/448b8a7c0007bf3ffaf7098b3363956401e82929/pillar/base/haproxy.sls#L140-L157). The "listens" directive effectively wires up a TCP proxy from haproxy to the backends.

JacobCoffee:
rebased against main, bugs comes up with the latest/greatest changes

pillar/base/haproxy.sls
```diff
@@ -67,6 +67,22 @@ haproxy:
       verify_host: planet.psf.io
       check: "HEAD / HTTP/1.1\\r\\nHost:\\ planet.psf.io"
 
+    {% for tracker, config in pillar["bugs"]["trackers"].items() %}
```
JacobCoffee, line 70: This needs fixin'

pillar/base/haproxy.sls
```diff
@@ -155,3 +171,16 @@ haproxy:
...
+
+    {% for (port, service, ssl) in [(25, "smtp", False), (587, "smtps", True), (465, "submission", True)] %}
+    roundup-{{ service }}:
+      bind: :{{ port }}
+      service: roundup-{{ service }}
+      extra:
+        - timeout client 30m
+        - timeout server 30m
+        {% if ssl %}
+        {# TODO: Need to update pillar data and haproxy.cfg.jinja to utilize tls #}
```
JacobCoffee, line 183: This needs implementing still

https://www.haproxy.com/documentation/haproxy-configuration-tutorials/ssl-tls/ says it needs a bind port

We already do 
```
{% for item in config.get("extra", []) -%}
{{ item }}
{% endfor -%}
```
So we will need special casing or need to move it out of the `extra` tree (?)

JacobCoffee: [5d93dbc](https://github.com/python/psf-salt/pull/393/commits/5d93dbc1a618575a16c8f0fdc5ba7d82bee22a63) aims to address this @ewdurbin

pillar/base/haproxy.sls
JacobCoffee: maybe just a 
```
include:
  - bugs
```

pillar/base/haproxy.sls
```diff
@@ -155,3 +171,12 @@ haproxy:
       extra:
         - timeout client 86400
         - timeout server 86400
+
+    {% for (port, service, ssl) in [(25, "smtp", False), (587, "smtps", True), (465, "submission", True)] %}
+    roundup-{{ service }}:
+      bind: :{{ port }} {% if ssl %} ssl crt /etc/ssl/private/bugs.psf.io.pem {% endif %}
```
ewdurbin, line 177: this cert won't exist on the haproxy host, did you mean to use bugs.python.org.pem?

pillar/base/haproxy.sls
ewdurbin: that's an option. we can also just write them out as we have for others like codespeed. matter of preference.

pillar/base/haproxy.sls
```diff
@@ -67,6 +67,22 @@ haproxy:
...
 
+    {% for tracker, config in pillar["bugs"]["trackers"].items() %}
+    {{ tracker }}:
+      domains:
+        - {{ config.server_name }}
+      verify_host: bugs.psf.io
+      check: "HEAD / HTTP/1.1\\r\\nHost:\\ {{ config.server_name }}"
+    {% endfor %}
+
+    pypy-web:
```
ewdurbin, line 78: looks like this returned-from dead in the rebase.

pillar/base/firewall/bugs.sls
```diff
@@ -9,3 +11,6 @@ firewall:
     port: 587
   submission:
     port: 465
+  frontend-bugs:
+    port: 9000
```
ewdurbin, line 15: do we need to open up 9001 and 9002 also?

salt/bugs/init.sls
```diff
@@ -4,6 +4,19 @@ include:
   - tls.lego
   - nginx
 
+/etc/consul.d/service-bugs.json:
```
ewdurbin, line 7: should this be moved inside the `for tracker` loop to register a service for each of the trackers, mapping to the name used in the haproxy config?

pillar/base/haproxy.sls
JacobCoffee: Yep 😭

JacobCoffee: [fd51660](https://github.com/python/psf-salt/pull/393/commits/fd51660a6642ac7bd8030aee06cb8a3d7eae66b5)

pillar/base/haproxy.sls
JacobCoffee: [578429e](https://github.com/python/psf-salt/pull/393/commits/578429eae608e309c550b7397e418f0e9dbae636)

salt/bugs/init.sls
JacobCoffee: [98a43e4](https://github.com/python/psf-salt/pull/393/commits/98a43e4435031274223870a854656f9229055ea3)

salt/bugs/init.sls
```diff
@@ -4,18 +4,20 @@ include:
   - tls.lego
   - nginx
 
-/etc/consul.d/service-bugs.json:
+{% for tracker, config in pillar["bugs"]["trackers"].items() %}
+/etc/consul.d/service-{{ tracker }}.json:
```
JacobCoffee, line 8: does this need to be service-{{tracker or just `{{ tracker }}` for full 1:1 mapping into haproxy?

JacobCoffee:
image
image
😕

ewdurbin (commented):
This is really close, I think hovering over the LastChk result will give more details on why they are red.

ewdurbin:
ya know what, @JacobCoffee, I'm realizing that smtps and submission may not even be properly configured on the bugs host. I think it may only be listening on `smtp` 🤦🏼 . the firewall holes for mail are shared, which I think led to the red herring. mail.python.org _does_ listen on those which is why it is included.

pillar/dev/top.sls
```diff
@@ -16,6 +16,7 @@ base:
   'bugs':
     - match: nodegroup
     - bugs
+    - firewall.bugs
```
ewdurbin, line 19: good catch!

salt/bugs/init.sls
```diff
@@ -210,7 +210,36 @@ postfix:
...
+        port: {{ port }}
+    - user: root
+    - group: root
+    - mode: "0644"
+    - require:
+      - pkg: consul-pkgs
+{% endfor %}
+
 {% for tracker, config in pillar["bugs"]["trackers"].items() %}
+/etc/consul.d/roundup-{{ tracker }}.json:
```
ewdurbin, line 230: just as a matter of convention, we generally put this state to create a service after the thing that exposes the service. in this case it is nginx. it's probably fine but it does keep things somewhat "linear" to read.

pillar/base/haproxy.sls
```diff
@@ -155,3 +163,12 @@ haproxy:
       extra:
         - timeout client 86400
         - timeout server 86400
+
+    {% for (port, service, ssl) in [(25, "smtp", False), (587, "smtps", True), (465, "submission", True)] %}
```
ewdurbin, line 167: remove unused services (smtps and submission) here?

ewdurbin (approved):
I think this is good to go! there's one lingering issue but it isn't technically a blocker.

salt/bugs/config/nginx.conf.jinja
```diff
@@ -77,3 +77,49 @@ server {
...
+  listen {{ port }} ssl;
+  server_name {{ server_name }};
+  include mime.types;
+
+  ssl_certificate /etc/ssl/private/bugs.psf.io.pem;
+  ssl_certificate_key /etc/ssl/private/bugs.psf.io.pem;
```
JacobCoffee, line 87:
```suggestion
  ssl_certificate /etc/ssl/private/bugs.python.org.pem;
  ssl_certificate_key /etc/ssl/private/bugs.python.org.pem;
```

JacobCoffee: @ewdurbin ?

ewdurbin: I don't think so. Since we are standing up _new_ server blocks for internal access on ports 9000-9002 and having them validate using the internal TLS, we want to use the internal TLS cert for bugs.psf.io.

JacobCoffee: 😵‍💫

ewdurbin (approved):

PR #400, opened 2024-07-22
chore(#399): remove `cia.vc` domain

### LINKED ISSUE #399: bugs: remove reference to stale domain
## What

There once was a service for git/IRC relays and we configured it at https://github.com/python/psf-salt/blob/bb73f8edbef6740445371b1ab9d6dacfd8369bd3/pillar/base/bugs.sls#L36.

This domain has been bought/repurposed and is now home to some sort of gambling ads.
https://scanlime.org/2011/05/cia-vc-service-is-down-indefinitely/

Remove the reference since we shouldnt be sending people searching through our codebase to that 😅

JacobCoffee:
## Description

- Renames the now stale URL for the retired cia.vc service. The domain has since been repurposed and is pointing to advertisements.

## Closes

- Closes #399

### DIFF
--- a/pillar/base/bugs.sls
+++ b/pillar/base/bugs.sls
@@ -32,7 +32,6 @@ bugs:
             spambayes_uri: "http://localhost:8001/sbrpc"
             spambayes_ham_cutoff: "0.2"
             spambayes_spam_cutoff: "0.85"
-            ciavc_server: "https://python.org"
       accept_email: False
       email_reject_message: "This tracker is in read-only mode. Please use GitHub issues to open a ticket: https://github.com/python/cpython/issues"
     jython:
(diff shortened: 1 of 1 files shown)


pillar/base/bugs.sls
```diff
@@ -32,7 +32,7 @@ bugs:
             spambayes_uri: "http://localhost:8001/sbrpc"
             spambayes_ham_cutoff: "0.2"
             spambayes_spam_cutoff: "0.85"
-            ciavc_server: "http://CIA.vc"
+            ciavc_server: "https://python.org"
```
JacobCoffee, line 35: My only reason for now just outright removing this is I am unsure if it will cause any breakage/outage or something not to load and spam logs.

ewdurbin: bet it is safe:

```
ee@bugs:/srv/roundup$ sudo grep -r 'ciavc_server' /srv/roundup/src
ee@bugs:/srv/roundup$ sudo grep -r 'ciavc' /srv/roundup/src
ee@bugs:/srv/roundup$ 
```

ewdurbin (approved):

pillar/base/bugs.sls
JacobCoffee: Good deal! Removing then

PR #432, opened 2024-08-01
fix(hg): adjust for duplicate key

JacobCoffee:
## Description

- Fixes duplicate key


JacobCoffee:
something is weird here because it is merged into main and git log on salt-master shows
image
that is is also there, but the `/src/salt/hg/init.sls` show the old file... and so deploying it does nothing

PR #447, opened 2024-08-07
feat(pythontest): add backup data housed under `srv/`

JacobCoffee:
## Description

- Adds backup state for `pythontest` that was missing


JacobCoffee:
https://github.com/python/pythontestdotnet

PR #462, opened 2024-08-14
feat(#449): add swap space by default

### LINKED ISSUE #449: Infra: Add swap space to hosts
## What

We need to add a small swap partition to hosts, especially needed for our hosts with minimal memory footprint.

1G is enough

JacobCoffee:
## Description

- Adds swap with configurability via pillar datum

https://serverfault.com/a/865797

## Closes

- Closes #449

### DIFF
--- a/salt/base/swap.sls
+++ b/salt/base/swap.sls
@@ -1,6 +1,6 @@
 {% set swapfile = salt["pillar.get"]("swapfile", {}) %}
 {% set size = swapfile.get("size", "1024") %}
-{% set path = swapfile.get("path", "/swapfile") %}
+{% set path = swapfile.get("path") %}
 
 {{ path }}:
   cmd.run:
--- a/salt/base/swap.sls
+++ b/salt/base/swap.sls
@@ -1,0 +1,16 @@
+{% set swapfile = salt["pillar.get"]("swapfile", {}) %}
+{% set size = swapfile.get("size", "1024") %}
+{% set path = swapfile.get("path", "/swapfile") %}
+
+{{ path }}:
+  cmd.run:
+    - name: |
+        swapon --show=NAME --noheadings | grep -q "^{{ path }}$" && swapoff {{ path }}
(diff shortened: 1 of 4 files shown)


JacobCoffee:
Currently blocking consul migration

salt/base/swap.sls
```diff
@@ -0,0 +1,16 @@
...
+    - name: |
+        swapon --show=NAME --noheadings | grep -q "^{{ path }}$" && swapoff {{ path }}
+        rm -f {{ path }}
+        fallocate -l {{ size }}M {{ path }}
+        chmod 0600 {{ path }}
+        mkswap {{ path }}
+    - unless: bash -c "[[ $(($(stat -c %s {{ path }}) / 1024**2)) = {{ size }} ]]"
+
+  mount.swap:
+    - persist: true
```
JacobCoffee, line 16: we could also add to fstab if that is better

salt/base/swap.sls
```diff
@@ -0,0 +1,16 @@
+{% set swapfile = salt["pillar.get"]("swapfile", {}) %}
+{% set size = swapfile.get("size", "1024") %}
+{% set path = swapfile.get("path", "/swapfile") %}
```
ewdurbin, line 3:
```suggestion
{% set path = swapfile.get("path") %}
```

I'm not sure want to _always_ add this (really thinking mostly in dev where it is almost never necessary), so give a way to opt out, see additional suggested change.

salt/base/swap.sls
```diff
@@ -0,0 +1,16 @@
...
+
+{{ path }}:
```
ewdurbin, line 5: Given suggested change above, wrapping this in a `{% if path is not None %}` would make this opt-in.

ewdurbin (approved):

salt/base/swap.sls
ewdurbin: this seems fine!

PR #464, opened 2024-08-14
feat: require swap to run before datadoge

JacobCoffee:
## Description

- Requires swap state to run before the doge
- Renames some things to be non-generic

### DIFF
--- a/salt/datadog/init.sls
+++ b/salt/datadog/init.sls
@@ -36,7 +36,7 @@ datadog-agent:
     - installed
     - require:
       - pkgrepo: datadog_repo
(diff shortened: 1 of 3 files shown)


salt/datadog/init.sls
```diff
@@ -36,6 +36,7 @@ datadog-agent:
     - installed
     - require:
       - pkgrepo: datadog_repo
+      - mount: {{ swap_path }}
```
ewdurbin, line 39: probably needs to be wrapped with the same if statement as the swap creation itself

ewdurbin (approved):

PR #466, opened 2024-08-14
fix: define swap vars in datadog state

JacobCoffee:
## Description

- Defines swap state vars in datadog state

### DIFF
--- a/salt/datadog/init.sls
+++ b/salt/datadog/init.sls
@@ -1,6 +1,4 @@
-{% set swap_file = salt["pillar.get"]("swap_file", {}) %}
-{% set swap_size = swap_file.get("swap_size", "1024") %}
-{% set swap_path = swap_file.get("swap_path") %}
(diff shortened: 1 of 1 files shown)


salt/datadog/init.sls
```diff
@@ -1,3 +1,7 @@
+{% set swap_file = salt["pillar.get"]("swap_file", {}) %}
+{% set swap_size = swap_file.get("swap_size", "1024") %}
+{% set swap_path = swap_file.get("swap_path") %}
```
ewdurbin, line 3:
```suggestion
{% set swap_path = salt["pillar.get"]("swap_file.swap_path") %}
```

not positive if this syntax is spot on, but this could simplify things to one line.

ewdurbin (approved):

salt/datadog/init.sls
JacobCoffee: I think its :

PR #472, opened 2024-08-18
use git not hg for PyPy

mattip:
## Description

PyPy has moved to git. This will make the logs show.

## Closes

See for instance https://speed.pypy.org/changes/?rev=a02837db72&exe=8&env=3 which fails to display. I no longer have permissions to access the live logs to see if this is part of the problem, but locally this fixes the page.

### DIFF
--- a/pillar/base/codespeed.sls
+++ b/pillar/base/codespeed.sls
@@ -25,7 +25,7 @@ codespeed-instances:
     module: speed_pypy
     wsgi_app: speed_pypy.wsgi:application
     clones:
-      git: {}
-      hg:
+      git:
         pypy:
-          source: https://foss.heptapod.net/pypy/pypy
+          source: https://github.com/pypy/pypy
+      hg: {}


ewdurbin (approved):

ewdurbin:
@mattip re "I no longer have permissions to access the live logs to see if this is part of the problem, but locally this fixes the page.", I'm not sure why that would be, your user with the keys [here](https://github.com/python/psf-salt/blob/37e194a5272ac259b7d756de37fb912a54f2d944/pillar/base/users.sls#L342-L350) grants you full access to the host running speed.pypy.org (codespeed.nyc1.psf.io)

ewdurbin:
This is applied but that URL still errors. Can you let me know if further assistance is needed @mattip ?

mattip:
Thanks.

I do have permissions, I just needed to remember to `sudo`.

The link works now. One problem was that I started uploading PyPy 3.10 results without properly setting the default branch to py3.10 on the new PyPy3.10 project, which needs to be done manually.

PR #476, opened 2024-08-22
fix(#474): auto generate keys per host

### LINKED ISSUE #474: new hg backups are failing
From cron to infrastructure-staff:

> Warning: Identity file /etc/backup/.ssh/id_rsa_hg-mercurial-static not accessible: No such file or directory.
[root@backup.sfo1.psf.io](mailto:root@backup.sfo1.psf.io): Permission denied (publickey).

and

> Warning: Identity file /etc/backup/.ssh/id_rsa_hg-svn-config not accessible: No such file or directory.
[root@backup.sfo1.psf.io](mailto:root@backup.sfo1.psf.io): Permission denied (publickey).

It looks like this is because we infer that a given backup config has an individual backup key when [templating the remote command](https://github.com/python/psf-salt/blob/37e194a5272ac259b7d756de37fb912a54f2d944/salt/backup/client/init.sls#L42-L44).

JacobCoffee:
## Description

- Automates key creation on each host for backups
- Publishes those keys' public key to the salt mines
- Retrieve those on the backup host and add them to authorized keys

## Closes

- Closes #474

### DIFF
--- a/salt/backup/server/init.sls
+++ b/salt/backup/server/init.sls
@@ -1,6 +1,14 @@
+{% set backup_hosts = salt['mine.get']('*', 'cmd.run') %}
+{% set target_host = salt['pillar.get']('backup:target_host') %}
 
 include:
   - backup.base
+
+/backup:
+  file.directory:
+    - user: root
+    - group: root
+    - mode: "0755"
 
 {% for backup, config in salt['pillar.get']('backup-server:backups', {}).items() %}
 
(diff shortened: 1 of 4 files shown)


salt/backup/server/init.sls
```diff
@@ -1,28 +1,50 @@
+{% set backup_hosts = salt['mine.get']('*', 'cmd.run') %}
+{% set target_host = salt['pillar.get']('backup:target_host') %}
```
JacobCoffee, line 2: should be specified as a `backup_target_host`

salt/backup/server/init.sls
```diff
@@ -1,28 +1,50 @@
...
     - name: {{ config['user'] }}
 
-{{ backup }}-ssh:
-  ssh_auth:
-    - present
+{{ backup }}-ssh-dir:
+  file.directory:
+    - name: /home/{{ config['user'] }}/.ssh
     - user: {{ config['user'] }}
-    - names:
```
JacobCoffee, line 15: wat

PR #477, opened 2024-08-22
fix(planet): remove lego in favor of haproxy

JacobCoffee:
## Description

- Finishes transfer of ssl cert handling to haproxy instead of lego

## Closes

- Closes python/planet#571

### DIFF
--- a/salt/planet/config/nginx.planet.conf.jinja
+++ b/salt/planet/config/nginx.planet.conf.jinja
@@ -19,8 +19,10 @@ server {
     server_name {{ site }};
     error_log /var/log/nginx/{{ site }}.error.log;
     access_log /var/log/nginx/{{ site }}.access.log;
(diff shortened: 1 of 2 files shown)


salt/planet/config/nginx.planet.conf.jinja
```diff
@@ -19,8 +19,10 @@ server {
     server_name {{ site }};
     error_log /var/log/nginx/{{ site }}.error.log;
     access_log /var/log/nginx/{{ site }}.access.log;
-    ssl_certificate /etc/lego/certificates/{{ grains['fqdn'] }}.crt;
```
ewdurbin, line 22: I think we can completely remove both of these server blocks (80, 443)

ewdurbin: Looking at current DNS we are still hitting the bugs host directly, so I don't think so.

ewdurbin (approved):

PR #486, opened 2024-08-28
Native haproxy consul integration

ewdurbin:
## Description

Migrates to using [consul DNS](https://developer.hashicorp.com/consul/docs/services/discovery/dns-static-lookups) via haproxy [server-template](https://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4.2-server-template) instead of consul-template for dynamic haproxy backends.

Drops consul-template as a dependency.

### DIFF
--- a/salt/haproxy/config/haproxy.cfg.jinja
+++ b/salt/haproxy/config/haproxy.cfg.jinja
@@ -1,6 +1,11 @@
 {% set haproxy = salt["pillar.get"]("haproxy", {}) -%}
 {% set psf_internal = salt["pillar.get"]("psf_internal_network") -%}
 
+resolvers consul
+    nameserver consul 127.0.0.1:8600
+    accepted_payload_size 8192
+    hold valid 5s
+
 global
     log /dev/log    local0
     log /dev/log    local1 notice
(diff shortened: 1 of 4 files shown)


salt/haproxy/config/haproxy.cfg.jinja
```diff
@@ -248,8 +252,7 @@ listen {{ name }}
     {{ line }}
     {% endfor %}
 
-    {{ "{{" }}range service "{{ config.service }}@{{ pillar.dc }}"}}
-    {% raw %}server {{.Node}} {{.Address}}:{{.Port}} check{{end}}{% endraw %}{% if config.get("send_proxy", False) %} send-proxy{% endif %}
+    server-template backend 1-3 _{{ config.service }}._tcp.service.{{ pillar.dc }}.consul resolvers consul resolve-opts allow-dup-ip resolve-prefer ipv4 check {% if config.get("send_proxy", False) %} send-proxy{% endif %}
```
JacobCoffee, line 255: is there a way to only template out 3 backends if we need them?

JacobCoffee (approved):

salt/haproxy/config/haproxy.cfg.jinja
ewdurbin: I suppose we could make the "count" tunable in service configuration. 3 was just "plucked from air", but since we only ever tend to use one backend, maybe 2 is a better default choice.

ewdurbin: Screenshot 2024-08-29 at 10 17 12 AM

ewdurbin: Implemented in 776fdd3, one drawback is that if we want multiple backends to be available during a migration we would need to ensure we update the "backends" value. Thoughts?

JacobCoffee: I think thats an acceptable thing, we just need docs on it. It will for sure unmuddy the LB dashboards 😬

JacobCoffee: but 2 is better than 3, so either way

ewdurbin: added some documentation in f15ab08

PR #490, opened 2024-09-08
Add docs access for Adam Turner

AA-Turner:
As discussed/suggested on the core dev discord (cc @hugovk).

A

PR #491, opened 2024-09-08
Order the users list alphabetically

AA-Turner:
Builds on top of #490 -- there seemed to be a vague ordering, so I added myself at the appropriate spot alphabetically, but when scrolling to the end that order evaporated -- this is a proposal to order the list of non-admin users.

To validate the changes in this PR, just reorder the keys in `users.sls` to have an alphabetical order.

A

### DIFF
--- a/pillar/base/users.sls
+++ b/pillar/base/users.sls
@@ -69,6 +69,45 @@ users:
           - docs
           - docsbuild
 
+  ambv:
+    fullname: "Łukasz Langa"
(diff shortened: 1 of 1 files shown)


ewdurbin:
Thanks for the PR. Thinking about this a bit more, I'm leaning towards refactoring so that each user is represented by its own file, what do you think about #491?

ewdurbin:
Closing in favor of #492

PR #493, opened 2024-09-17
Update rdiff-backup command.

msapiro:
I'm seeing messages with content "WARNING: this command line interface is deprecated and will disappear, start using the new one as described with '--new --help'." from mail.python.org when cron runs /usr/local/backup/mail-python-org/scripts/backup.bash.

I'm not at all sure that this change is the appropriate fix, but I think so.

## Description

-

## Closes

-


ewdurbin:
Downgrading rdiff-backup client on the host resolved the issue.

It had to be manually upgraded to a version compatible with the newest backup server, but got upgraded too far.

PR #496, opened 2024-09-23
Redirect dead batteries

AA-Turner:
Previous mass removals of modules have had redirects, we can do the same for PEP 594:

https://github.com/python/psf-salt/blob/20e0021e000b8ca670def1d7e9ebfdaefa64ff4a/salt/docs/config/nginx.docs-backend.conf#L53-L55

Note we can't have `3|` yet -- needs to be added post 3.13.0 release.

A


JacobCoffee (approved):

AA-Turner:
Thanks - I missed the modules already removed in 3.12.

A

JacobCoffee (approved):

PR #498, opened 2024-09-25
Redirect full version archive files to ``{major}.{minor}``

AA-Turner:


salt/docs/config/nginx.docs-backend.conf
```diff
@@ -200,6 +200,11 @@ server {
         }
     }
 
+    # Map archive URLs to {major}.{minor}
+    location ~ ^/([a-z-]*/)?(3|3.12|3.13|3.14)/archives/python-(3.12|3.13|3.14)\.\d+((a|b|rc)\d)?-docs(-(html|pdf-a4|pdf-letter|texinfo|text)\.(zip|tar\.bz2)|\.epub)$ {
+        return 301 https://$host/$1$2/archives/python-$3-docs$6;
+    }
```
AA-Turner, line 206: This might be clearer?: 

cc @JacobCoffee -- I don't know if you have or know of any guidance on best formatting for these redirect rules.
```suggestion
    # Map archive URLs to {major}.{minor}
    location ~ ^/([a-z-]*/)?(3|3.12|3.13|3.14)/archives/python-(3.12|3.13|3.14)\.\d+((a|b|rc)\d)?-docs-(html|pdf-a4|pdf-letter|texinfo|text)\.(zip|tar\.bz2)$ {
        return 301 https://$host/$1$2/archives/python-$3-docs-$6.$7;
    }
    location ~ ^/([a-z-]*/)?(3|3.12|3.13|3.14)/archives/python-(3.12|3.13|3.14)\.\d+((a|b|rc)\d)?-docs\.epub$ {
        return 301 https://$host/$1$2/archives/python-$3-docs.epub;
    }
```

JacobCoffee: given the nature of the capture groups i think there isnt much formatting that could save us! it looks pretty good to me, imo

I *do* think the separation for epub just adds a whole bunch of duplication and could still sit on the original single location block with the rest... but maybe @ewdurbin has more insight

ewdurbin:
For the sake of understanding... Can you give a few explicit examples of what the redirects desired are? I'm having a hard time working backward from the regex

AA-Turner:
* https://docs.python.org/pt-br/3.13/archives/python-3.13.0rc2-docs-pdf-a4.zip -> https://docs.python.org/pt-br/3.13/archives/python-3.13-docs-pdf-a4.zip
* https://docs.python.org/3.13/archives/python-3.13.0rc2-docs-pdf-a4.zip -> https://docs.python.org/3.13/archives/python-3.13-docs-pdf-a4.zip

The full list of expected filenames is in the opening post, redirecting a `{release}` version to a `{major}.{minor}`. This is applied over the full range of possible prefixes on d.p.o (languages & versions).

A

merwok:
Are these redirects effective _if_ the original URL is not found as a file?
Otherwise I don’t see why redirect a specific release doc to an imprecise one (I may be confused after having just read multiple tickets about docs builds)

AA-Turner:
The redirects are always active regardless of the filesystem -- the issue is that an archive called "3.X.Y" is not the same as the documentation at time of Python 3.X.Y's release. It is simpler to explain when we just present the archives for Python 3.X, and state that they're the current state of the branch -- to get versioned archives, use https://docs.python.org/release/.

A

PR #510, opened 2024-10-09
Use Surrogate-Key headers for docs.python.org

### LINKED ISSUE #287: Add some VCL to fastly so docs can be purged by top or 2nd level folder
As documented in https://docs.fastly.com/en/guides/wildcard-purges

It would be nice to be able to purge a whole version of the doc at once from docsubild scripts when updating a symlink, like `PURGE /3/` and `PURGE /fr/3/` and so on, instead of doing it file by file.

AA-Turner:
Closes #287.

A

### DIFF
--- a/salt/docs/config/nginx.docs-redirects.conf
+++ b/salt/docs/config/nginx.docs-redirects.conf
@@ -182,9 +182,19 @@ location ~ ^/(searchindex.js|objects.inv)$ {
 # Redirect ``$request_uri`` -> ``$request_uri.html``,
 # where the latter is a valid webpage.
 location ~ ^/((2|3)(\.[0-9]+)?|dev)/\w+/[\d\w\.]+(?!\.html)$ {
+    add_header Surrogate-Key en/$1 always;
     if (-f "${request_filename}.html") {
         return 301 https://$host:$request_uri.html;
     }
+}
+
+# Add the Surrogate-Key for bulk CDN purging
+# xref https://docs.fastly.com/en/guides/working-with-surrogate-keys
+location ~ ^/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key en/$1 always;
+}
+location ~ ^/(es|fr|id|it|ja|ko|pl|pt-br|tr|uk|zh-cn|zh-tw)/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key $1/$2 always;
 }
 
 # Map old, 2.5-and-earlier directory names to 2.6-and-later names.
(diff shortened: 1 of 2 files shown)


salt/docs/config/nginx.docs-redirects.conf
```diff
@@ -182,11 +182,21 @@ location ~ ^/(searchindex.js|objects.inv)$ {
...
 }
 
+# Add the Surrogate-Key for bulk CDN purging
+# xref https://docs.fastly.com/en/guides/working-with-surrogate-keys
+location ~ ^/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key en/$1 always;
+}
+location ~ ^/(es|fr|id|it|ja|ko|pl|pt-br|tr|uk|zh-cn|zh-tw)/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key $1/$2 always;
+}
```
AA-Turner, line 198: Ideally I had hoped to use some sort of non-binding match construct to add the headers, but it seems that NGINX `location` directives will always bind. The end goal is simply to add some unique prefix for every second level path (and English) that we can use to purge. I think the approach here is fine given every other rule here is a redirect, but worth pointing out.

Fastly [has some text](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/purging/#adding-surrogate-keys) about adding surrogate keys via `vcl_fetch`, but I don't know where I'd configure that.

A

salt/docs/config/nginx.docs-redirects.conf
```diff
@@ -185,6 +185,16 @@ location ~ ^/((2|3)(\.[0-9]+)?|dev)/\w+/[\d\w\.]+(?!\.html)$ {
...
+    add_header Surrogate-Key en/$1 always;
+}
+
+# Add the Surrogate-Key for bulk CDN purging
+# xref https://docs.fastly.com/en/guides/working-with-surrogate-keys
+location ~ ^/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key en/$1 always;
+}
+location ~ ^/(es|fr|id|it|ja|ko|pl|pt-br|tr|uk|zh-cn|zh-tw)/((2|3)(\.[0-9]+)?|dev)/ {
+    add_header Surrogate-Key $1/$2 always;
```
AA-Turner, line 197: Ideally I had hoped to use some sort of non-binding match construct to add the headers, but it seems that NGINX `location` directives will always bind. The end goal is simply to add some unique prefix for every second level path (and English) that we can use to purge. I think the approach here is fine given every other rule here is a redirect, but worth pointing out.

Fastly [has some text](https://www.fastly.com/documentation/guides/concepts/edge-state/cache/purging/#adding-surrogate-keys) about adding surrogate keys via `vcl_fetch`, but I don't know where I'd configure that.

A

ewdurbin (approved):

salt/docs/config/nginx.docs-redirects.conf
ewdurbin: This could be configured in our Fastly service for docs.python.org, but is probably more scrutable (and testable, nice!) here.

PR #511, opened 2024-10-09
ci: give write permissions to reminder

JacobCoffee:
## Description

- Deployminder sometimes fails due to permissions I think
https://github.com/python/psf-salt/actions/runs/11259227771/job/31307708334?pr=510


ewdurbin:
Interesting... why would they succeed sometimes without it?

ewdurbin (approved):

JacobCoffee:
@ewdurbin that is my confusion here as well :\

ewdurbin:
Doesn't appear that this worked

PR #515, opened 2024-10-10
ci: remove perms

JacobCoffee:
## Description

- remove perms from deployminder after testing a bunch on a private repo.

### DIFF
--- a/.github/workflows/deployminder.yml
+++ b/.github/workflows/deployminder.yml
@@ -11,8 +11,6 @@ jobs:
   remind:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
-    permissions:
-        issues: write
     steps:
       - name: Comment on PR
         uses: actions/github-script@v7


ewdurbin (approved):

.github/workflows/deployminder.yml
```diff
@@ -11,8 +11,6 @@ jobs:
   remind:
     if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
-    permissions:
-        issues: write
     steps:
       - name: Comment on PR
         uses: actions/github-script@v7
```
JacobCoffee, line 16: maybe should revert to v6

ewdurbin: ah, was it working before update?

ewdurbin: Doesn't look like we were ever using v6?

JacobCoffee: im not sure anymore :\ this works on a private repo as many times as i can trigger it, but here it fails. both repos have the same perms in settings > actions, so im just confused on what the issue is.

it was working before the permissions add, then it wasn't

JacobCoffee: going off of https://github.com/actions/github-script?tab=readme-ov-file#comment-on-an-issue and https://github.com/litestar-org/litestar/blob/119169d1c335f68cbc4673006b4cb2aa96983513/.github/workflows/pr-merged.yml#L49

PR #518, opened 2024-10-11
Avoid multiple redirects in pre-2.5 paths

AA-Turner:
This flattens multiple redirects in the Python 2.5 redirects.

We also remove redirects for `/dist/` and `/inst/` as `/distutils` and `/install` have been removed as pages on the website.

A


JacobCoffee (approved):

hugovk:
We got a report at https://github.com/python/pythondotorg/issues/2646 that http://docs.python.org/install/ and http://docs.python.org/distutils/ are 404s.

They were 404s before the redirects were removed here, but because they are evidently still linked to, shall we add redirects to the last existing versions?

Namely:

* https://docs.python.org/3.11/install/index.html
* https://docs.python.org/3.11/distutils/index.html

PR #525, opened 2024-11-09
Add syntax highlighting to NGINX config files

shenanigansd:
Tiny QoL improvement
I was linked to https://github.com/python/psf-salt/blob/main/salt/docs/config/nginx.docs-redirects.conf but there were no colors.
I've added a `.gitattributes` to prompt GitHub to apply the NGINX syntax highlighting.

ref: https://github.com/github-linguist/linguist/blob/main/docs/overrides.md
ref: https://github.com/github-linguist/linguist/blob/main/lib/linguist/languages.yml#L4712-L4727


AA-Turner (approved):
Thank you Bradley!

A

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Add syntax highlighting to NGINX config files
+nginx.*.conf linguist-language=Nginx
```
zware, line 2: I'm not sure how many wildcards this will take, but making it `nginx*.conf*` would also catch [nginx.conf.jinja](https://github.com/python/psf-salt/blob/d25f5bf1e001fe2bc8ebe59ea0ee2d24fcf4846a/salt/buildbot/config/nginx.conf.jinja), for example.

zware (approved):
Looks better to me :)

.gitattributes
JacobCoffee: would that make us lose Jinja highlighting?

zware: I don't think so, because AFAICT we don't currently have *any* highlighting there :)

zware: Ah, I did find a counterexample in [haproxy/config/nginx-redirect.conf.jinja](https://github.com/python/psf-salt/blob/d25f5bf1e001fe2bc8ebe59ea0ee2d24fcf4846a/salt/haproxy/config/nginx-redirect.conf.jinja), so maybe my suggestion is not a good one (though it would not actually catch that particular one).

PR #526, opened 2024-11-12
Add tests for the 'deprecations' redirect

AA-Turner:
re #522.

A


AA-Turner:
Odd that GHA hasn't yet started...

PR #542, opened 2024-12-18
chore: increase rate limit to not be so aggressive

JacobCoffee:
## Description

- Increase the rate limit for roundup trackers to not be so aggressive

### DIFF
--- a/pillar/base/haproxy.sls
+++ b/pillar/base/haproxy.sls
@@ -75,7 +75,7 @@ haproxy:
         - {{ config.server_name }}
       verify_host: bugs.psf.io
       check: "HEAD / HTTP/1.1\\r\\nHost:\\ {{ config.server_name }}"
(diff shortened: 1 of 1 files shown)


pillar/base/haproxy.sls
```diff
@@ -75,7 +75,7 @@ haproxy:
         - {{ config.server_name }}
       verify_host: bugs.psf.io
       check: "HEAD / HTTP/1.1\\r\\nHost:\\ {{ config.server_name }}"
-      rate_limit: {{ config.get('rate_limit', 10) }}
+      rate_limit: {{ config.get('rate_limit', 100) }}
```
ewdurbin, line 78: I’m nearly certain roundup can’t handle 100 rpm, and I _think_ we’ll see effectively 2x the rate since the counters aren’t shared between haproxy instances. So it’s not clear if this will result in any significant defense.

PR #543, opened 2024-12-18
remove separate tls listen for haproxy

ewdurbin:
haproxy has natively supported http/2 for a _long_ time.

This listen (combined frontend/backend) was added long ago before haproxy supported http/2 (back when it was called spdy).

Drop the http server config for http/2, and move all binds back into the main frontend.


ewdurbin:
basically a long overdue revert of https://github.com/python/psf-salt/commit/794c3cc5f430cb297bde5ab067bcc7627e6fabf9

JacobCoffee (approved):
`The ssl_fc function returns true if the connection came in over HTTPS` neat

PR #552, opened 2025-02-25
Add moin access to `adam.sls`

AA-Turner:
Following up on the proposed wiki work from the most recent documentation meeting, we're exploring a `git` replica of the wiki content. Better to do this via the data dump than taxing the already-overloaded web server.

### DIFF
--- a/pillar/base/users/adam.sls
+++ b/pillar/base/users/adam.sls
@@ -7,6 +7,11 @@ users:
         - docs
         - docsbuild
         sudo: true
+      moin:
+        allowed: true
+        groups:
+          - moin
+        sudo: true
     fullname: Adam Turner
     ssh_keys:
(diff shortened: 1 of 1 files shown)


JacobCoffee:
@AA-Turner Could you bring me up to speed on the wiki work or point me at some meeting minutes perhaps?

JacobCoffee (approved):

AA-Turner:
@JacobCoffee Thanks! See https://docs-community.readthedocs.io/en/latest/monthly-meeting/2025-02.html, there's also some discussion in the docs discord forum. Currently exploratory work, but the idea is having a single source for our documentation -- moving useful things from the wiki to docs.python.org & then redirecting users to d.p.o, as it has a much more clearly defined contribtion & review process (and much more transparent).

A

AA-Turner:
```
❯ ssh adam@moin.nyc1.psf.io
adam@moin.nyc1.psf.io: Permission denied (publickey).
```

I assume there's a wait period for config to deploy?

PR #557, opened 2025-03-17
chore: add pubkey

JacobCoffee:
## Description

- new key


ewdurbin (approved):
given that Jacob has full admin access, validated request out-of-band via dm/selfie.

PR #558, opened 2025-03-18
feat: starttls

### LINKED ISSUE #556: Add support for SMTPS on roundup-tracker.org
Currently roundup-tracker.org users use email heavily. Due to a recent personal issue, I have had to move
to gmail as my email client.

I am unable to send email to issues@roundup-tracker.org because gmail wants to use starttls to send it over
a secure channel (probably port 25 since 587 isn't in use on the server). I get the following error from gmail:
```
The response from the remote server was:

454 4.7.0 TLS not available due to local problem 
```
in a delivery status notification (delay) from the gmail mail daemon.

It looks like this error is addressed in: https://www.24x7serversecurity.com/blog/unable-to-send-mail-using-postfix-starttls-authentication-4-7-0-tls-not-available-due-to-local-problem/

Would it be possible to enable startls so gmail users can correspond with roundup?

For testing, trying to open a ticket using gmail should work. You will get a delay message in 4 or so hours
on gmail if it can't establish a TLS connection.

Thanks.

-- rouilj

JacobCoffee:
## Description

- Enables optional (may) starttls for receiving (smptd)

## Closes

- Closes #556

### DIFF
--- a/salt/bugs/config/postfix/main.cf
+++ b/salt/bugs/config/postfix/main.cf
@@ -24,8 +24,8 @@ compatibility_level = 3.6
 
 
 # TLS parameters
-smtpd_tls_cert_file=ssl_certificate /etc/ssl/private/bugs.psf.io.pem;
-smtpd_tls_key_file=etc/ssl/private/bugs.psf.io.pem;
+smtpd_tls_cert_file=/etc/ssl/private/bugs.psf.io.pem
+smtpd_tls_key_file=/etc/ssl/private/bugs.psf.io.pem
 smtpd_tls_security_level=may
 
 smtp_tls_CApath=/etc/ssl/certs


salt/bugs/config/postfix/main.cf
```diff
@@ -24,8 +24,8 @@ compatibility_level = 3.6
 
 
 # TLS parameters
-smtpd_tls_cert_file=ssl_certificate /etc/ssl/private/bugs.psf.io.pem;
-smtpd_tls_key_file=etc/ssl/private/bugs.psf.io.pem;
+smtpd_tls_cert_file=/etc/ssl/private/bugs.psf.io.pem
```
ewdurbin, line 27: Will this certificate be trusted by clients?

JacobCoffee:
in [e655530](https://github.com/python/psf-salt/pull/558/commits/e6555308bcfbf9339a13548bd50b4060e422a39e) pebble is running

```
vagrant@salt-master:~$ curl -k https://salt-master.vagrant.psf.io:14000/dir
{
   "keyChange": "https://salt-master.vagrant.psf.io:14000/rollover-account-key",
   "meta": {
      "externalAccountRequired": false,
      "termsOfService": "data:text/plain,Do%20what%20thou%20wilt"
   },
   "newAccount": "https://salt-master.vagrant.psf.io:14000/sign-me-up",
   "newNonce": "https://salt-master.vagrant.psf.io:14000/nonce-plz",
   "newOrder": "https://salt-master.vagrant.psf.io:14000/order-plz",
   "revokeCert": "https://salt-master.vagrant.psf.io:14000/revoke-cert"
```
but hitting:
```
2025-04-01 15:15:14,982:ERROR:certbot._internal.log:An unexpected error occurred:
2025-04-01 15:15:14,982:ERROR:certbot._internal.log:requests.exceptions.SSLError: HTTPSConnectionPool(host='salt-master.vagrant.psf.io', port=14000): Max retries exceeded with url: /dir (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)')))
```

salt/tls/init.sls
```diff
@@ -25,3 +31,101 @@ ssl-cert:
...
+
+# initial test
+{% if grains['id'] == 'salt.nyc1.psf.io' or grains['id'] == 'salt-master.vagrant.psf.io' %}
+pypa.io:
+  acme.cert:
+    - email: infrastructure-staff@python.org
+    - webroot: /etc/lego
+    - renew: 14
+    {% if pillar["dc"] == "vagrant" %}
+    - server: https://salt-master.vagrant.psf.io:14000/dir
```
ewdurbin, line 43: docs say there's built in support for using a test CA: with `test_cert`, which is probably the easiest way? https://docs.saltproject.io/en/3006/ref/states/all/salt.states.acme.html

ewdurbin: Ah, I see that just uses Lets Encrypt's test server, which still tries to validate.

ewdurbin: @JacobCoffee 0aefff26ea07e16abfa97cb90237eb862e6487cd should unblock you.

salt/tls/init.sls
```diff
@@ -25,3 +44,91 @@ ssl-cert:
     - require:
       - pkg: ssl-cert
 {% endfor %}
+
+{% if salt["match.compound"](pillar["roles"]["salt-master"]["pattern"]) %}
+# HTTP-validated domains
+{% for domain in [
+    'pypa.io',
```
ewdurbin, line 51: Lists like these are always a good sign that something should be in pillar rather than a state.

I'd suggest moving this list into a subkey of https://github.com/python/psf-salt/blob/7fcd4a36b615cb629f3afa60912f0b128b99e4f9/pillar/base/tls.sls#L4

Maybe `acme_certs`.

Something like:

```
tls:
  acme_certs:
    example.com:
      validation: http
      roles:
      - loadbalancer
      - example
      additional_sans:
      - www.example.com
```

This would also simplify the logic in your pillar extension to allow us to determine where certs are validated (once dns is supported) and distributed.

salt/_extensions/pillar/ca.py
```diff
@@ -295,46 +296,203 @@ def get_ca_signed_cert(cacert_path, ca_name, CN):
...
+    Reads ACME certificates and determines which ones should be available
+    to the specified minion based on access rules.
+    """
+    acme_certs = {}
+    
+    try:
+        print(f"Processing ACME certificates for minion: {minion_id}")
+        all_acme_certs = _find_acme_certs()
+        
+        # Check if this is a loadbalancer (gets all certs)
```
ewdurbin, line 410: This is all a lot of work, when we could theoretically just assign certs explicitly to roles, see other comment.

salt/_extensions/pillar/ca.py
```diff
@@ -299,42 +299,52 @@ def ext_pillar(minion_id, pillar, base="/etc/ssl", name="PSFCA", cert_opts=None)
...
             # Add the signed certificates to the pillar data
             cert_data = get_ca_signed_cert(base, name, certificate)
             data["tls"]["certs"][certificate] = cert_data
 
+    # Collect ACME certs (acme.cert) for this minion based on its roles
+    acme_certs = pillar.get("tls", {}).get("acme_certs", {})
+    for domain, domain_config in acme_certs.items():
+        cert_roles = domain_config.get("roles", [])
+        if any(role in minion_roles for role in cert_roles):
+            data["tls"]["acme_certs"][domain] = domain_config
```
ewdurbin, line 348: i'd have thought that this would be populating the actual certificate data into pillar so the minion can use it.

pillar/base/tls.sls
```diff
@@ -61,3 +61,77 @@ tls:
     svn.psf.io:
       roles:
         - hg
+
+  acme_certs:
+    bugs.python.org:
```
ewdurbin, line 66: I don't see a reason to have a separate instance here just for bugs.python.org. The same cert can be used for the load balancer and the roundup box.

salt/bugs/config/postfix/main.cf
ewdurbin, line 27: This would now be `/etc/ssl/private/acme-bugs.python.org.pem` right?

pillar/base/tls.sls
```diff
@@ -61,3 +61,73 @@ tls:
...
+        - www.jython.net
+        - jython.net
+        - www.jython.com
+        - jython.com
+    bugs.python.org:
+      validation: http
+      name: bugs.python.org
+      roles:
+        - loadbalancer
+        - bugs
```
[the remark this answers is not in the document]
ewdurbin: ope spun up docs to check that also... and we're distorting these to all roles/hosts:

```
vagrant@docs:~$ sudo ls /etc/ssl/private/
acme-bugs.python.org.pem   acme-pycon.org.pem		acme-speed.pypy.org.pem  PLACEHOLDER.pem
acme-jython.org.pem	   acme-pypa.io.pem		dhparams.pem		 ssl-cert-snakeoil.key
acme-planetpython.org.pem  acme-salt-public.psf.io.pem	docs.psf.io.pem
```

salt/bugs/config/postfix/main.cf
JacobCoffee: [0da94f1](https://github.com/python/psf-salt/pull/558/commits/0da94f14147c0b1f2556b4a7a7a7214e53ea6bbe)

PR #560, opened 2025-04-11
Use the correct spelling of the ``--languages`` flag

AA-Turner:
docsbuild-scripts expects the languages flag to be ``--languages``, not ``--language``. We want to remove the singular variant, so this updates the usage to allow removal upstream.

A


ewdurbin (approved):

ewdurbin:
Will go live on the next */15

AA-Turner:
Thanks!

PR #561, opened 2025-04-11
Docsbuild config

ewdurbin:
## Description

Make environment variables specific to docs.python.org available as a configuration file to ensure they're used no matter how the script is invoked.

Note: This limits access to the file to user/group `docsbuild`

Ref: https://github.com/python/docsbuild-scripts/pull/269
Ref: https://github.com/python/docsbuild-scripts/pull/266

### DIFF
--- a/salt/docs/init.sls
+++ b/salt/docs/init.sls
@@ -67,29 +67,17 @@ virtualenv-dependencies:
     - onchanges:
       - git: docsbuild-scripts
 
-docsbuild-analytics:
-  cron.env_present:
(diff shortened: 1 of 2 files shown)


salt/docs/init.sls
```diff
@@ -67,29 +67,17 @@ virtualenv-dependencies:
     - onchanges:
       - git: docsbuild-scripts
 
-docsbuild-analytics:
```
ewdurbin, line 70: These cron.env_present will have to be removed manually after https://github.com/python/docsbuild-scripts/pull/269 is live.

PR #566, opened 2025-04-29
feat: manage moin host apache2 conf

JacobCoffee:
## Description

- source hosts apache2 conf
- add request timings to log format
- manage file via salt


salt/moin/configs/moin.apache.conf.jinja
```diff
@@ -0,0 +1,125 @@
...
+        Options Indexes FollowSymLinks
+        AllowOverride None
+        Require all granted
+</Directory>
+
+#<Directory /srv/>
+#       Options Indexes FollowSymLinks
+#       AllowOverride None
+#       Require all granted
+#</Directory>
```
JacobCoffee, line 87:
```suggestion
```

salt/moin/configs/moin.apache.conf.jinja
```diff
@@ -108,7 +108,7 @@ AccessFileName .htaccess
 # Note that the use of %{X-Forwarded-For}i instead of %h is not recommended.
 # Use mod_remoteip instead.
 LogFormat "%v:%p %h %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"" vhost_combined
-LogFormat "%h %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"" combined
+LogFormat "%h %l %u %t \"%r\" %>s %O %D \"%{Referer}i\" \"%{User-Agent}i\"" combined
```
ewdurbin, line 106: oh! hm I don't think I followed when we were chatting earlier.

we shouldn't need to template this entire file, I think a new LogFormat can be specified in the moin file itself and then set just for that location.

```
LogFormat "%h %l %u %t \"%r\" %>s %O %D \"%{Referer}i\" \"%{User-Agent}i\"" psf_moin
CustomLog /var/log/apache2/wiki.access.log psf_moin
```

can go [here](https://github.com/python/psf-salt/blob/e9d84fd04b6cf5115ab0a81d682a35127737f1cc/salt/moin/configs/wiki.python.org.conf#L20C5-L20C56) and do this in a 2 line diff without overwriting the root apache conf.

ewdurbin (changes requested):

ewdurbin (approved):

PR #572, opened 2025-05-07
fix planet job, can't use tty if there's no tty

ewdurbin:


ewdurbin:
fixes https://github.com/python/planet/issues/597

PR #573, opened 2025-05-07
buildbot: Use IPv4 address rather than 'localhost'

zware:
Buildbot does not listen on the IPv6 interface.

## Description

-

## Closes

-


zware:
That was the fastest merge I've ever seen.

PR #585, opened 2025-06-03
feat: migrate from `deadmanswitch` to Sentry monitors

JacobCoffee:
## Description

- Swap to Sentry cron monitors

https://python-software-foundation.sentry.io/issues/alerts/rules/?project=4509407540346880

The only downside is that org tokens have a limited scope (I think only `org:ci` currently) and we need `alerts:write/read` and `project:read`... so this currently is using my token and in prod the same would need to happen.


pillar/base/sentry.sls
```diff
@@ -0,0 +1,6 @@
+sentry:
```
ewdurbin, line 1: Feels weird that these are all in base, and that all are in clear-text.

It _appears_ to me that everything here is sufficient for someone to report check-ins on our behalf, including [updating the interval](https://docs.sentry.io/product/crons/getting-started/http/#creating-or-updating-a-monitor-through-a-check-in-optional).

salt/_extensions/pillar/sentry_cron.py
```diff
@@ -0,0 +1,119 @@
...
+            print(f"Failed to create monitor: {e}")
+            return {}
+
+        if monitor.status_code == 201:
+            monitor_id = monitor.json()["id"]
+            with tempfile.NamedTemporaryFile('w', dir=str(base_path), delete=False) as tf:
+                tf.write(monitor_id)
+                tempname = tf.name
+            os.replace(tempname, minion_path)
+            print(f"Created monitor with ID: {monitor_id}")
```
ewdurbin, line 115: it's unclear to me how we use the stored monitor ids.

pillar/base/sentry.sls
JacobCoffee: this makes it more interesting i missed that you could check in on a non existent monitor!

JacobCoffee: wdyt about [1456a65](https://github.com/python/psf-salt/pull/585/commits/1456a6540f403be2ceff8c11ae9555e96db87b9c)

salt/base/auto-highstate.sls
```diff
@@ -1,18 +1,24 @@
...
-  cron.present:
-    - identifier: 15m-interval-highstate
-    - name: "timeout 5m salt-call state.highstate >> /var/log/salt/cron-highstate.log 2>&1; curl https://nosnch.in/{{ dms_token }} &> /dev/null"
-    - minute: '*/15'
-{% else %}
 15m-interval-highstate:
   cron.present:
     - identifier: 15m-interval-highstate
-    - name: "timeout 5m salt-call state.highstate >> /var/log/salt/cron-highstate.log 2>&1"
+    - name: {{ '/usr/local/bin/sentry-checkin.sh' if sentry_enabled else 'timeout 5m salt-call state.highstate >> /var/log/salt/cron-highstate.log 2>&1' }}
```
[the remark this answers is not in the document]
JacobCoffee: i think i did this as part of [1456a65](https://github.com/python/psf-salt/pull/585/commits/1456a6540f403be2ceff8c11ae9555e96db87b9c)

JacobCoffee:
todo:
- how to create pagertduty alert from failures
- or just send slack message

pillar/dev/secrets/sentry.sls
```diff
@@ -0,0 +1,5 @@
+secrets:
```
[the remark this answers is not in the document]
ewdurbin: No def keep these!!!!

JacobCoffee, line 1: we dont need all this nesting i think , 
like how `datadog/init.sls` has `api_key: {{ pillar.get('datadog_api_key') }}`

salt/base/auto-highstate.sls
```diff
@@ -1,18 +1,24 @@
-{% set dms_token = salt["pillar.get"]("deadmanssnitch:token") %}
+{% set sentry_enabled = salt["pillar.get"]("secrets:sentry:project_id") and salt["pillar.get"]("secrets:sentry:project_key") and salt["pillar.get"]("secrets:sentry:ingest_url") %}
+
+curl:
```
ewdurbin, line 3: should this and the next clause be conditional on `sentry_enabled` also?

salt/base/scripts/sentry-checkin.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+MINION_ID=$(salt-call --local grains.get id --out=newline_values_only)
```
ewdurbin, line 3: Should all of these be templated verbatim or do we need to do `salt-call` each time? Seems we could just use the value from pillar/pillar_ext when we write/update the file.

JacobCoffee: [130b807](https://github.com/python/psf-salt/pull/585/commits/130b8073359ae81259a615fb46399115b3d57d13)

pillar/dev/top.sls
```diff
@@ -8,6 +8,7 @@ base:
     - tls
     - users.*
     - postgres.clusters
+    - secrets.sentry
```
JacobCoffee, line 11: @ewdurbin would we need this really since we dont expect it to run in dev and the "do we run this?" checks are if the keys in pillar secrets exist

ewdurbin: Yeah, no need to _always_ have it on. Maybe just comment this line out so people have an idea where to turn it on.

PR #586, opened 2025-06-05
Add diegorusso to the list of users

diegorusso:
I've also given myself access to codespeed as in the past I worked with it and very likely I will need it in the near future.


diegorusso:
@ambv FYI

PR #589, opened 2025-06-10
chore: format for better editing

JacobCoffee:
## Description

- per https://develop.sentry.dev/sdk/telemetry/check-ins/#monitor-upsert-support, adds `failure_issue_threshold` so we only create an issue (and therefore a page) after 4 misses, also resolves an issue after 1 successful checkin


JacobCoffee:
image
successful changes after deploy

PR #597, opened 2025-08-01
fix: update to prevent NO_PUBKEY 64CBBC8173D76B3F

JacobCoffee:
## Description

-

## Closes

-


ewdurbin:
Can you add a little bit of a narrative here about what is happening? I feel like we churn on these strings/keys a lot if we could make sure this is the last straw that would rule.

PR #598, opened 2025-08-04
Add redirects for new translations

AA-Turner:
## Description

Bengali, Greek, and Romanian are new translations. Tests updated.


AA-Turner:
@JacobCoffee is this PR live? I'm not seeing the redirects locally, but might be due to caching etc.

A

PR #600, opened 2025-08-07
Reduce frequency of non-HTML documentation builds

AA-Turner:
We currently build ~every 48 hours. This changes the non-HTML (PDF, EPUB, TexInfo, Text) documentation builds to be every nine days, or roughly thrice per month.

A


JacobCoffee:
very nice! i'll get this out into production when im back at my computer

PR #603, opened 2025-08-11
Add Savannah for downloads and docs server access

savannahostrowski:
## Description

Related to https://discuss.python.org/t/welcome-the-3-16-and-3-17-release-manager-savannah-bailey/100163! Looks like things have changed a bit since Hugo last did this in https://github.com/python/psf-salt/pull/356 but I think this is correct.


JacobCoffee (approved):

JacobCoffee:
i'll get this out on the morrow

PR #604, opened 2025-08-15
Increase frequency of English HTML documentation builds

AA-Turner:
This increases the frequency of checking for new updates to rebuild for the English HTML-only documentation builds to once every five minutes (offset by one minute).

When there is nothing to do (no changes), the build job exits in less than 3 seconds. When there is work to do (a new documentation change in CPython), a full rebuild takes around three minutes.

Increasing the frequency of rebuilds for HTML-only English should therefore not put significant extra load on the server, whilst also pushing out updates faster after a change is made.

A

### DIFF
--- a/salt/docs/init.sls
+++ b/salt/docs/init.sls
@@ -110,7 +110,7 @@ docsbuild-only-html:
 
 docsbuild-only-html-en:
   cron.present:
-    # run twice hourly at HH:16 and HH:46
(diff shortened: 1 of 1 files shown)


JacobCoffee:
Should be rolling out shortly @AA-Turner

AA-Turner:
Thanks @JacobCoffee!

PR #607, opened 2025-09-09
Update buildbot service to Python 3.13

zware:


salt/buildbot/init.sls
```diff
@@ -70,10 +69,12 @@ update-master:
   cmd.run:
     - runas: buildbot
     - cwd: /srv/buildbot
-    - name: make update-master
+    - name: make update-master PYTHON_VERSION=3.13
```
zware, line 72: Is there a handy way to set a "buildbot_python_version" variable somewhere and use it here and in the package names above so that there's just one place to update?

salt/buildbot/init.sls
```diff
@@ -70,10 +69,12 @@ update-master:
...
     - require:
+      - pkg: buildbot-deps
       - git: /srv/buildbot
     - onchanges:
+      - pkg: buildbot-deps
```
zware, line 77: I'm unsure if this is specified correctly, but my goal is to make sure this is run after this is merged (and after any future updates).  Do both `require` and `onchanges` need to be set?

zware:
Great, thanks!

PR #640, opened 2026-04-03
Fix `error_page` for versioned doc subdirectory URLs

StanFromIreland:
For example, currently https://docs.python.org/3.14/whatsnew/3.1t4 doesn’t work. Why: I missed a `location` block :-(

Also, add some tests, which would have been a good idea before.

## Description

-

## Closes

-


JacobCoffee (approved):

StanFromIreland:
Thanks Jacob!

PR #647, opened 2026-08-14
Add a Dependabot config file

StanFromIreland:
Along with a few little miscellaneous changes. We're running some pretty outdated Actions. I'll Zizmor it in a follow up PR.


StanFromIreland:
Little ping @JacobCoffee in case you missed this.

PR #648, opened 2026-08-18
Add PEP 545 redirect for Persian translation

sepehr-rs:
## Description
This PR adds PEP 545 redirects for Persian translation

PR #650, opened 2026-09-10
Add nginx redirect for German translation

sba72:
Adds the `de` language code to the docs redirect rules, following the addition of German to the language switcher.

### DIFF
--- a/salt/docs/config/nginx.docs-redirects.conf
+++ b/salt/docs/config/nginx.docs-redirects.conf
@@ -8,7 +8,7 @@ location = / {
 }
 
 # Python 3 docs are the default at the root of each translation.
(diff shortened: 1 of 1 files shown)


JacobCoffee (approved):

StanFromIreland:
Oh, this was a little premature. We haven't added it to the switcher yet (https://github.com/python/docsbuild-scripts/pull/355).

StanFromIreland:
And this is missing all of the tests? @sba72 can you please open a PR with them, like we have in https://github.com/python/psf-salt/pull/648.

hugovk:
Tests added in https://github.com/python/psf-salt/pull/654.

PR #653, opened 2026-09-13
Build docs for security branches

### LINKED ISSUE #433: `run_release` should remind the release manager to start rebuild of security-only branches full docs on docs server
As noted [here](https://github.com/python/docsbuild-scripts#manually-rebuild-a-branch), a rebuild of the full set of docs for security-only branches is not run automatically on the docs build server as is done for bugfix and feature branches. This was done to substantially reduce the CPU load on the docs build server since the security-only branches do not receive many doc updates and releases are as needed.

Because of this, there should be a RM reminder added to `run_release.py` to ensure that the docs build update has been or will be manually started for security branch releases. (That step should be added to PEP 101 as well if it isn't there already - I didn't see it.)

hugovk:
Draft until https://github.com/python/docsbuild-scripts/pull/356 is merged.

## Description

- Currently docs for security branches must be build manually. We often forget to do it after security releases. Let's include in the cron:

  * Nothing extra into the 5-minute English HTML cron, so that stays fast.
  * The English + translations HTML as part of the daily cron.
  * The English + translations non-HTML in the 3-4 times/month cron.

- Docs/translations for security branches aren't often updated, so I don't expect this to cause too much extra load.
- If it does cause too much load, we can adjust the crons so security are built less often (or easily disable).

## Closes

- Closes https://github.com/python/release-tools/issues/433.


hugovk:
https://github.com/python/docsbuild-scripts/pull/356 is merged, ready for review.

salt/docs/init.sls
StanFromIreland, line 100: I'm a little worried this is slightly too often for security branches, could we do just every week?

hugovk: I think we might be okay with daily?

Some rough numbers suggest we might have an extra 3 builds per day taking an extra 20 minutes, or with a weekly cron with 10 extra builds taking an extra 70 minutes.

But let's measure and see how it is.

We can download logs (I've downloaded a fresh batch) and summarise build times like:
https://github.com/python/docsbuild-scripts/blob/4e3feeffb39c32524d80633e831395fa1ca15e6e/check_times.py#L8-L10

JacobCoffee (approved):

salt/docs/init.sls
StanFromIreland: Ok, let's try it.

StanFromIreland (approved):

